Issue #15, opened 2023-07-25
[Backend] As a User, my uploaded keywords are processed immediately


Goose97:
Prerequisites: #13 #14

## Why

Processing upload keywords immediately provides a smooth and snappier interface, enhances the application UX

## Acceptance Criteria

- Upload keywords are converted to Sidekiq jobs and enqueued immediately
- Sidekiq jobs are distributed to workers and processed immediately (the keyword processing logic will be handled by #14)
- Retry mechanism MUST be implemented
- If a job exceeds maximum retry times, it should be abandon. The status is updated in the database accordingly

PR #20, opened 2023-07-27
[Backend] Scrape keyword data from the Google Search page

### LINKED ISSUE #14: [Backend] Scrape keyword data from the Google Search page
## Why

Scraping keywords data  is our core logic

## Acceptance Criteria

- Given a keyword, it should answer:
  - ~~Number of AdWords advertisers in the top position~~
  - ~~Total number of AdWords advertisers on the page~~
  - ~~URLs of the AdWords advertisers in the top position~~
  - ~~Number of the non-AdWords results on the page~~
  - ~~URLs of the non-AdWords results on the page~~
  - All search entries on the page. Each entry includes its kind (ads / non_ads), position (top / bottom / nil) and urls (one entry can have multiple urls)
  - Total number of links (all of them) on the page
  - HTML code of the page/cache of the page
- All results MUST be persisted in the database

Goose97:
Closes [[Backend] Scrape keyword data from the Google Search page](https://github.com/Goose97/google-scraper-ruby/issues/14)

## What happened 👀

- Implement Google search service
- Implement Google search result parse service
- Implement scrape service 

## Insight 📝

- Use Nokogiri for HTML parsing
- Use Faraday as HTTP client. It offers great utilities to perform mocking during testing

## Proof Of Work 📹

All tests are passed

Screenshot 2023-07-27 at 21 22 23
Screenshot 2023-07-27 at 21 22 48
Screenshot 2023-07-27 at 21 22 10


app/commands/scrape_command.rb
```diff
@@ -0,0 +1,47 @@
...
+
+class ScrapeCommand
+  attr_reader :keyword
+
+  def initialize(keyword_id:)
+    @keyword_id = keyword_id
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
```
Goose97: I'm not sure how to address these two lint rules. IMO, this method is simple enough to a human to parse/understand.

I would prefer not to extract the rescue part to separate functions as it fragments the method.

.rubocop.yml
```diff
@@ -21,6 +21,9 @@ AllCops:
 Style/Documentation:
   Enabled: false
 
+Style/NegatedIf:
+  Enabled: false
```
olivierobert: Let's avoid making changes to Rubocop as it defeats the purpose of having shared code style conventions. Adapting the code to fit the team's code style is best.

.rubocop.yml
```diff
@@ -50,3 +53,6 @@ RSpec/NestedGroups:
 
 RSpec/MultipleExpectations:
   Max: 10
+
+RSpec/ExampleLength:
+  Max: 10
```
olivierobert, line 55: Same here. It might a sign that tests might need to be broken down.

Gemfile
```diff
@@ -11,6 +11,8 @@ gem 'discard' # Soft deletes for ActiveRecord
 gem 'sidekiq' # background processing for Ruby
 gem 'bootsnap', require: false # Reduces boot times through caching; required in config/boot.rb
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby] # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
+gem 'faraday' # HTTP client
+gem 'nokogiri' # Gem to work with XML and HTML files
```
olivierobert: The descriptions can be improved slightly:

```suggestion
gem 'farday' # HTTP client library, with support for multiple backends
gem 'nokogiri' # Work with XML and HTML from Ruby
```

> **Note**
> In general, you can take the short description on the repository
> image

app/commands/scrape_command.rb
olivierobert: The method is too long because of the double `rescue` block and logging statements. While it is readable, there is complexity in the error handling, that's the code smell that Rubocop is raising. Using `rescue` blocks should be limited to maximum one (these are not comparable to functional paradigms like `case/else` or `with/else` in Elixir). To achieve this, it is best to raise one type of error, not multiple.

app/commands/scrape_command.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+class ScrapeCommand
```
olivierobert: While there is nothing wrong with command objects, it is best to restrict yourself to the default objects listed in the project structure: https://nimblehq.co/compass/development/code-conventions/ruby/ruby-on-rails/#project-structure. The team's convention is to have only service objects, so if a service can call other objects, e.g., `ScrapService` can call the `Search` and `Parse` services. 

If the command pattern is something you want to introduce to the team, you can do it. You must follow [the change management guidelines](https://nimblehq.co/compass/intro/change-management/). It will likely more time though, due to the team sizes to get an agreement on this before we can merge this PR. So it is better to stick to the current conventions.

app/commands/scrape_command.rb
```diff
@@ -0,0 +1,47 @@
...
+    - error: #{e}"
+
+    raise e
+  end
+  # rubocop:enable Metrics/MethodLength
+  # rubocop:enable Metrics/AbcSize
+
+  private
+
+  def persist_scrape_result(keyword, parse_result)
```
olivierobert: A more idionatic Ruby method name would be:

```suggestion
  def save_scrape_result(keyword, parse_result)
```

As it is consistent with the ActiveRecord API.

app/commands/scrape_command.rb
```diff
@@ -0,0 +1,47 @@
...
+    @keyword_id = keyword_id
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
+  def run
+    keyword = Keyword.find(@keyword_id)
+    html = Google::SearchService.search!(keyword.content)
+    result = Google::ParseService.new(html).call
+    persist_scrape_result keyword, result
```
olivierobert: Prefer to add a lin break between variable statements and method calls for readability:

```suggestion

    persist_scrape_result keyword, result
```

Make the code breathe 🌬️

app/models/keyword.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Keyword < ApplicationRecord
+  enum status: { processing: 'processing', succeeded: 'succeeded', failed: 'failed' }
+  has_many :keyword_search_entries, dependent: :destroy
+  validates :content, :status, presence: true
```
olivierobert, line 8: Again here, line break issues:

```suggestion
	enum status: { processing: 'processing', succeeded: 'succeeded', failed: 'failed' }
	
  has_many :keyword_search_entries, dependent: :destroy
  
  validates :content, :status, presence: true
```

Make sure to follow this structure for all models: https://nimblehq.co/compass/development/code-conventions/ruby/ruby-on-rails/#models

app/services/google/parse_service.rb
```diff
@@ -0,0 +1,65 @@
...
+        search_entries: [*top_ads, *bottom_ads, *non_ads_normal, *non_ads_video],
+        links_count: @doc.css('a[href]').count,
+        result_page_html: @html
+      )
+    end
+
+    private
+
+    def top_ads
+      @doc.css('#taw div[data-text-ad]').map do |ad_div|
```
olivierobert: Avoid hard-coding HTML selectors in method calls. For code sicoverabaility, it would be best to define these as constants.

olivierobert: We can avoid using `@` to access instance variables by defining these variables as private `attr_accessor`:

```
private

attr_accessor :html, :doc
```

Then it becomes:

```suggestion
      doc.css('#taw div[data-text-ad]').map do |ad_div|
```

It is an idiomatic Ruby way 🧑‍🏫

app/services/google/parse_service.rb
```diff
@@ -0,0 +1,65 @@
...
+      @doc.css('#search div[data-vurl][jsaction]').map do |div|
+        SearchEntry.new(
+          kind: :non_ads,
+          urls: extract_links(div),
+          position: nil
+        )
+      end
+    end
+
+    def extract_links(doc) = doc.css('a[href][data-ved]').pluck(:href)
```
olivierobert: Single-line methods are rarely used in Ruby (again, it is different from Elixir 😉 )

app/services/google/search_service.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Google
+  module SearchService
+    SearchServiceError = GoogleScraperRuby::Errors::SearchServiceError
+    HOST = 'www.google.com'
+    # rubocop:disable Layout/LineLength
+    USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36'
```
olivierobert: To prevent being blocked when doing multiple searches, you will need to implement user agent rotation. So it is best to start planning adding this in this PR.

app/services/google/search_service.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Google
+  module SearchService
+    SearchServiceError = GoogleScraperRuby::Errors::SearchServiceError
```
olivierobert: This error muest be defined in: https://github.com/Goose97/google-scraper-ruby/pull/20/files#diff-abc06f343e143ef5f7301a4a974abcace033c43ef32b083a833a91e87a11def9

app/services/google/search_service.rb
```diff
@@ -0,0 +1,32 @@
...
+        raise SearchServiceError.new(uri,
+                                     "Response with status code #{response.status}")
+      end
+
+      response.body
+    rescue Faraday::ConnectionFailed, Faraday::ServerError, Faraday::ClientError => e
+      raise SearchServiceError.new uri, e
+    end
+
+    def self.build_conn(uri)
```
olivierobert: `conn` is something taken from Phoenix which has no place in Rails 😅 Here the method name should be:

```suggestion
    def http_client
```

Since it is not used elsewhere, it could be a private instance method in a class.

app/services/google/search_service.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Google
+  module SearchService
```
olivierobert: Modules with method are rarely used. It is not idiomatic in Ruby unless for mixins. So this rather be a standard class. In Rails, it is best to follow OOP patterns, are the framework itself follows them. Other Ruby frameworks follow functional programming oriented paradigms but Rails is not one of them,

spec/commands/scrape_command_spec.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+SearchServiceError = GoogleScraperRuby::Errors::SearchServiceError
+
+RSpec.describe ScrapeCommand, type: :service do
+  describe '#run' do
+    context 'given an existing keyword_id' do
+      it 'updates the scrape result to the database if the command succeeds', vcr: 'google/google_no_ads' do
```
olivierobert: Prefern waraping conditional into a context to adoid lengthy example descriptions:

```ruby
context 'given an existing keyword' do
  context 'given NO errors' do
    it 'save the scraped result', vcr: 'google/google_no_ads' do

    end
  end

  context 'given errors' do
    it 'raises an error' do

    end

    it 'logs the error' do

    end
  end
end
```

> **Note**
> Notice the capitalization of `NO` to make it easy at first glance to know what the test is about.

spec/services/google/parse_service_spec.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Google::ParseService, type: :service do
+  describe '#call' do
+    context 'when parsing a page having 3 top ads and 1 bottom ad' do
```
olivierobert: The test descriptions are too coupled with the fixtures making it hard to maintain. Prefer the following structure:

```
context 'given a page with top and bottom ads' do
  it 'retrieves the top ads', vcr: 'google/programming_courses_3_top_1_bottom' do

  end

  it 'retrieves the bottom ads', vcr: 'google/programming_courses_3_top_1_bottom' do

  end
end
```

spec/services/google/parse_service_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+
+      it 'counts 4 ads in total', vcr: 'google/programming_courses_3_top_1_bottom' do
+        html = Google::SearchService.search!('programming courses')
+        bottom_ads = described_class.new(html).call.search_entries.count { |e| e.kind == :ads }
+        expect(bottom_ads).to eq(4)
+      end
+    end
+
+    context 'when parsing a page having no ads' do
```
olivierobert: An improved description would be:

```suggestion
    context 'a page WITHOUT any ads' do
```

spec/services/google/parse_service_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      it 'counts 0 ads', vcr: 'google/google_no_ads' do
+        html = Google::SearchService.search!('google')
+        ads = described_class.new(html).call.search_entries.count { |e| e.kind == :ads }
+        expect(ads).to eq(0)
+      end
+    end
+
+    context 'when parsing a page contains an ad for apple.com/store' do
```
olivierobert: These tests are duplicated with the ones for top and bottom contexts. See if these can be merged.

spec/services/google/search_service_spec.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Google::SearchService, type: :service do
+  let(:stubs) { Faraday::Adapter::Test::Stubs.new }
+  let(:stubbed_conn) { Faraday.new { |b| b.adapter(:test, stubs) } }
+
+  describe '#search!' do
+    before(:each) do
```
olivierobert: Avoid using let and before blocks as much as possible: https://nimblehq.co/compass/development/code-conventions/ruby/rspec/#convention-15. There is added complexity when nesting context with the order of execution which can cause unexpected side effects. Prefer to inline the mock in each test even if it comes at the cost of repetitions, and define helper methods instead.

olivierobert (changes requested):

app/services/google/search_service.rb
Goose97: well, this is just an alias, for better readability (again, inspired by Elixir)

app/services/google/search_service.rb
Goose97: actually, it's Faraday's terminology (https://www.rubydoc.info/github/lostisland/faraday/Faraday.new)

However, I admit that a framework-agnostic name would be better. Good point.

app/services/google/search_service.rb
Goose97: got it. While we're at it, which one is more idiomatic:

Using class methods
```ruby
module Google
  class SearchService
    def search!(keyword)
    end
  end
end
```

Using instance methods
```ruby
module Google
  class SearchService
    def initialize(keyword)
      @keyword = keyword
    end

    def search!
    end
  end
end
```

Gemfile
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/802acb5ec77fc85f179ae4c144dd321054e4e1a5

.rubocop.yml
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/7818449712a3ebb39bbdc369b1e88555cdf9ef4c

.rubocop.yml
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/7818449712a3ebb39bbdc369b1e88555cdf9ef4c

app/commands/scrape_command.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/defa7f47009b8c127be165fef468c28c5b83d6d9

app/commands/scrape_command.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/defa7f47009b8c127be165fef468c28c5b83d6d9

app/models/keyword.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/b2aef75f8792a54884596c0bd2e63c805121f4ff

app/services/google/parse_service.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits

app/services/google/search_service.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/2f3c3930c6f5b0f8a32041d9e8e88ae9b7e91828

app/commands/scrape_command.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/8756759ccba2f72dc834b4e0b82d835307bbacda

spec/services/google/parse_service_spec.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/72b6f0e445d318a8a42cf6cddd88123223a0ef1b#diff-a64e6d1a295f3b99057c1fa3f636d0b6ce962ebb7c5957bbda89e526ab28795bL35-L36

spec/services/google/parse_service_spec.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/9988cc8d24c0bff170a67f8cc38e7b31a6adf90d

spec/services/google/parse_service_spec.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/72b6f0e445d318a8a42cf6cddd88123223a0ef1b

spec/commands/scrape_command_spec.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/642b749eead67d6bae83f506917819c2f03c55ca

spec/services/google/search_service_spec.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/a65b875f9864c4458482e66e048d38338ae16bad

However, this also makes some test cases exceed the spec example length limit. IMO, we can disable the lint rule in these cases

app/services/google/parse_service.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    TOP_ADS_SELECTOR = '#taw div[data-text-ad]'
+    BOTTOM_ADS_SELECTOR = '#bottomads div[data-text-ad]'
+    NON_ADS_NORMAL_SELECTOR = '#search div[data-snhf]'
+    NON_ADS_VIDEO_SELECTOR = '#search div[data-vurl][jsaction]'
+    GENERAL_LINKS_SELECTOR = 'a[href]'
+    INSIDE_SEARCH_ENTRY_LINKS_SELECTOR = 'a[href][data-ved]'
+
+    private_constant :TOP_ADS_SELECTOR, :BOTTOM_ADS_SELECTOR, :NON_ADS_NORMAL_SELECTOR, :NON_ADS_VIDEO_SELECTOR,
+                     :GENERAL_LINKS_SELECTOR, :INSIDE_SEARCH_ENTRY_LINKS_SELECTOR
```
olivierobert, line 28: Using 	`private_contant` is quite rare (i.e., rarely seen in Ruby codebases) 🤔  These could be moved in the public interface without any issues, but it is fine to keep as-is.

app/services/google/search_service.rb
olivierobert: Ah I see, but again this is Ruby 😅 Aliases are not a thing in Ruby. So you need to follow Ruby idiomatic ways. You cannot/shouldn't code Ruby following another language's code style.

app/services/google/search_service.rb
```diff
@@ -0,0 +1,32 @@
...
+        raise SearchServiceError.new(uri,
+                                     "Response with status code #{response.status}")
+      end
+
+      response.body
+    rescue Faraday::ConnectionFailed, Faraday::ServerError, Faraday::ClientError => e
+      raise SearchServiceError.new uri, e
+    end
+
+    def self.http_client(uri)
```
olivierobert: This does not need to be a class method:

```suggestion
    private 
    
    def http_client(uri)
```

app/services/google/search_service.rb
```diff
@@ -0,0 +1,32 @@
...
+    SearchServiceError = GoogleScraperRuby::Errors::SearchServiceError
+    HOST = 'www.google.com'
+    # rubocop:disable Layout/LineLength
+    USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36'
+    # rubocop:enable Layout/LineLength
+
+    def self.search!(keyword)
+      uri = URI::HTTPS.build(host: HOST, path: '/search', query: { q: keyword }.to_query)
+      conn = http_client uri
+      response = conn.get(uri, {}, { 'User-Agent' => USER_AGENT })
```
olivierobert: Code style improvemens: add blank lines, rename variables to complete names and more consistency in using parentheses (you used them for `URI::HTTPS.build` but not for the line below)

```suggestion
      uri = URI::HTTPS.build(host: HOST, path: '/search', query: { q: keyword }.to_query)
      client = http_client(uri)
      response = client.get(uri, {}, { 'User-Agent' => USER_AGENT })
      
```

app/services/google/search_service.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Google
+  class SearchService
+    SearchServiceError = GoogleScraperRuby::Errors::SearchServiceError
+    HOST = 'www.google.com'
+    # rubocop:disable Layout/LineLength
+    USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36'
+    # rubocop:enable Layout/LineLength
+
+    def self.search!(keyword)
```
olivierobert: Since it is class, let's use a constructor method and an instance method to follow OOP:

```ruby
 def initialize(keyword)
   @keyword = keyword
end

def search!
   #...
end

private

attr_reader :keyword
```

app/services/google/search_service.rb
olivierobert: Will this be addressed in this PR or another one?

spec/services/google/parse_service_spec.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Google::ParseService, type: :service do
+  describe '#call' do
+    context 'given a page with top and bottom ads' do
+      it 'retrieves the top ads', vcr: 'google/programming_courses_3_top_1_bottom' do
+        html = Google::SearchService.search!('programming courses')
+        top_ads = described_class.new(html).call.search_entries.count { |e| e.kind == :ads && e.position == :top }
```
olivierobert: You seem to like compacting code 😅 It hurts readability for humans. Since all tests usually follow the [AAA pattern](https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80), it helps to add a line break around each block.

```ruby
html = Google::SearchService.search!('programming courses')

top_ads = described_class.new(html).call.search_entries.count { |e| e.kind == :ads && e.position == :top }

expect(top_ads).to eq(3)
```

spec/services/google/parse_service_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        expect(top_ads).to eq(3)
+      end
+
+      it 'retrieves the bottom ads', vcr: 'google/programming_courses_3_top_1_bottom' do
+        html = Google::SearchService.search!('programming courses')
+        bottom_ads = described_class.new(html).call.search_entries.count { |e| e.kind == :ads && e.position == :bottom }
+        expect(bottom_ads).to eq(1)
+      end
+
+      it 'retrieves all the ads', vcr: 'google/programming_courses_3_top_1_bottom' do
```
olivierobert, line 58: This test is kind of redundant with the two previous tests. It does not increase confidence. See [the test matrix](https://nimblehq.co/compass/development/testing/#test-matrix) to understand more the tradeoff about confidence when it comes to testing 🧑‍🏫

spec/services/google/parse_service_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    context 'given a page WITHOUT any ads' do
+      it 'counts 0 ads', vcr: 'google/google_no_ads' do
+        html = Google::SearchService.search!('google')
+        ads = described_class.new(html).call.search_entries.count { |e| e.kind == :ads }
+        expect(ads).to eq(0)
+      end
+    end
+
+    context 'given a page with non-ads entries' do
```
olivierobert: How is this different from the previous context? You might misunderstand the purpose of `context` as the goal is not to create blocks but identify the different branches of code. In this case, there are only two branches (page with or without ads), so only 2 contexts are required.

spec/services/google/scrape_service_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        it 'raises an error' do
+          allow(Google::SearchService).to receive(:search!).and_raise(SearchServiceError, '')
+          allow(Rails.logger).to receive(:error)
+
+          expect do
+            described_class.new(keyword_id: Fabricate(:keyword).id).call
+          end.to raise_error(SearchServiceError)
+        end
+
+        it 'logs an error' do
```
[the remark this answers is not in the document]
olivierobert: @Goose97 That is fine to increase the line count to `8` or `10` as you previously wanted to do it.

spec/services/google/scrape_service_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+          begin
+            described_class.new(keyword_id: Fabricate(:keyword).id).call
+          rescue SearchServiceError
+            expect(Rails.logger).to have_received(:error).with(/error while processing command/)
+          end
+        end
+      end
+    end
+
+    context 'given an non-existing keyword_id' do
```
olivierobert: English syntax issue and rewording:

```suggestion
    context 'given an INVALID keyword_id' do
```

spec/services/google/scrape_service_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        command = described_class.new(keyword_id: 42)
+        allow(Rails.logger).to receive(:error)
+
+        expect do
+          command.call
+        end.to raise_error(ActiveRecord::RecordNotFound)
+      end
+
+      it 'logs an error' do
+        command = described_class.new(keyword_id: 42)
```
olivierobert: There is no more command object anymore so this needs to be renamed to match the current implementatio

app/services/google/search_service.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/3d1e46eaf9eb5a2136fcad7ebdf5189fb0e49c28

spec/services/google/search_service_spec.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'support/http_client_helpers'
+
+RSpec.describe Google::SearchService, type: :service do
+  describe '#search!' do
+    context 'when search a keyword' do
+      it 'perform a request against www.google.com' do
```
olivierobert: Again, some issues with the usage of context here. For consistency, prefer using `given` (instead of mixing `given` and `when` in the same codebase). Here there should be two contexts: `given a successful request` and `given a failed request`

spec/services/google/search_service_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        HttpClientHelpers.stub_http_adapter do |adapter, connection|
+          allow(described_class).to receive(:http_client).and_return(connection)
+          adapter.get('/search') do |env|
+            outer_env = env
+            200
+          end
+        end
+
+        described_class.search!('hello world')
+        expect(outer_env.url.host).to include('www.google.com')
```
olivierobert: Many of the tests in this file test implementation details, which is a code smell. Instead, only test the behavior (the B in BDD testing). The tests should not know too much about the implementation but only care about the output of the code. It makes tests easier to maintain as we can swap the implementation while keeping the same behavior without breaking the tests. 

So to illustrate this point, there is no value in checking which host the code requests to. What matters is what the request returns. So here for simplicity we can also use `vcr` wand not use any stubs.

olivierobert (changes requested):

spec/services/google/parse_service_spec.rb
Goose97: good catch, I just missed this one 😰

spec/services/google/parse_service_spec.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/3c9ab33336160ac9cae55cb2a98d4440075c95ee

Goose97: The initial rationale is to make sure parse service returns exclusively :top and :bottom ads. Now I think about it, it's neither a feature requirement nor this test ensures that.

spec/services/google/parse_service_spec.rb
Goose97: Fixed in f4450f1df3ade108e9cc56259f081318d4caa2d3

spec/services/google/scrape_service_spec.rb
Goose97: Fixed in c8f0111f6aedc64645a3299d7e6d9034e5afa174

spec/services/google/search_service_spec.rb
Goose97: I agree with your opinion regarding the target host test case

To make sure we're on the same page. This test case
```ruby
it 'raise GoogleScraperRuby::Errors::SearchServiceError if the request returns a 4xx status code' do
  HttpClientHelpers.stub_http_adapter do |adapter, connection|
    allow(described_class).to receive(:http_client).and_return(connection)
    adapter.get('/search') { [429, {}, 'Too Many Requests'] }
  end

  expect do
    described_class.search!('hello world')
  end.to raise_error(GoogleScraperRuby::Errors::SearchServiceError)
end
```

also knows a lot about SearchService implementation (the face that it calls `http_client` to build the http client). So a better solution is to use a vcr which returns a 429 response. This will make sure the test is still correct even if we change the HTTP client library. Is it correct?

Goose97: also, I'm struggling to decide if this is a behavior we want to test:

```ruby
it "raise GoogleScraperRuby::Errors::SearchServiceError if it can't connect to the host" do
  HttpClientHelpers.stub_http_adapter do |adapter, connection|
    allow(described_class).to receive(:http_client).and_return(connection)
    adapter.get('/search') { raise Faraday::ConnectionFailed }
  end

  expect do
    described_class.search!('hello world')
  end.to raise_error(GoogleScraperRuby::Errors::SearchServiceError)
end
```

If it's worth testing then how should we do it. The current implementation is leaking implementation details i.e.  the `Faraday::ConnectionFailed` error

app/services/google/search_service.rb
Goose97: Fixed in 86cae6015bd328076e65b0f3887b828090f28c13

spec/services/google/search_service_spec.rb
Goose97: Fixed in df8db9fa717be526c4f37e066016095bc23c30e6

app/lib/google_scraper_ruby/errors.rb
```diff
@@ -0,0 +1,33 @@
...
+          - original_error: #{original_error}"
+        ERR_MESSAGE
+      end
+
+      private
+
+      attr_reader :original_error, :url
+    end
+
+      def initialize(url, error = nil)
```
olivierobert: Seems like a rebase or copy/paste gone wrong as this code block seems to come from somewhere unexpected

olivierobert (changes requested):

spec/services/google/parse_service_spec.rb
Goose97: I'm indeed misunderstand the meaning of the context block.

So in this case, should this test

```ruby
context 'given any page' do
  it 'includes the html of such page', vcr: 'google/monitor_50_links' do
    html = Google::SearchService.search!('monitor')
    expect(described_class.new(html).call.result_page_html).to be_a String
  end
end
```

appear in both contexts?

.rubocop.yml
andyduong1920: @Goose97 please double check as the `RSpec/ExampleLength:` is still here while it should be removed by the [7818449](https://github.com/Goose97/google-scraper-ruby/commit/7818449712a3ebb39bbdc369b1e88555cdf9ef4c), is there something wrong in the git flow?

app/services/google/parse_service.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    attr_accessor :doc, :html
```
andyduong1920: How about we just allow read only?

```suggestion
    attr_reader :doc, :html
```

app/services/google/parse_service.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def top_ads
+      doc.css(TOP_ADS_SELECTOR).map do |ad_div|
+        SearchEntry.new(
```
andyduong1920: instead of using the SearchEntryStruct here, how about we use the ActiveModel record directly @Goose97 as we are rarely use/define raw Struct in a Rails project, mostly we will use ActiveRecord or ActiveModel for that.

So here how about we use `KeyWordSearchEntry.new(xxx)`

andyduong1920: If you plan to use the Struct here btw, I would suggest to add the Struct Suffix to make it clear

app/models/keyword_search_entry.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class KeywordSearchEntry < ApplicationRecord
+  enum kind: { ads: 'ads', non_ads: 'non_ads' }
+  enum position: { top: 'top', bottom: 'bottom' }
```
andyduong1920: btw we have the rightads position as well @Goose97 

image

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Google
+  class ScrapeService
+    attr_reader :keyword
```
andyduong1920: Is it keyword or keyword_id 🤔 

```suggestion
    attr_reader :keyword_id
```

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def initialize(keyword_id:)
+      @keyword_id = keyword_id
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/AbcSize
+    def call
+      keyword = Keyword.find(@keyword_id)
```
andyduong1920: if we defined `attr_reader :keyword_id` above, here we just use `keyword_id` without the `@`

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,50 @@
...
+  class ScrapeService
+    attr_reader :keyword
+
+    def initialize(keyword_id:)
+      @keyword_id = keyword_id
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/AbcSize
+    def call
```
andyduong1920: as this method could raise the exception, let follow the Ruby convention 

- use `!` suffix if the method could raise exception

https://albertoalmagro.com/three-things-you-must-know-about-ruby-bang-methods/


```suggestion
    def call!
```

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,50 @@
...
+      Rails.logger.error "[#{self.class.name}]: keyword doesn't exist
+    - keyword_id: #{@keyword_id}"
+
+      raise e
+    rescue GoogleScraperRuby::Errors::SearchServiceError => e
+      Rails.logger.error "[#{self.class.name}]: error while processing request
+    - keyword: #{keyword.content}
+    - error: #{e}"
+
+      raise e
```
andyduong1920: Shall we update the `failed` status for the Keyword here?

app/services/google/parse_service.rb
```diff
@@ -0,0 +1,79 @@
...
+        )
+      end
+    end
+
+    def non_ads_normal
+      doc.css(NON_ADS_NORMAL_SELECTOR).map do |div|
+        SearchEntry.new(
+          kind: :non_ads,
+          urls: extract_links(div),
+          position: nil
```
andyduong1920: instead of `nil` can we use `page` or `main_search` to make it clear @Goose97 🤔

db/migrate/20230727072004_create_keywords.rb
```diff
@@ -0,0 +1,26 @@
+class CreateKeywords < ActiveRecord::Migration[7.0]
+  def change
+    create_enum  :keyword_status, %w[processing succeeded failed]
+
+    create_table :keywords, id: :serial do |t|
+      t.string :content, limit: 255, null: false
+      t.enum :status, enum_type: 'keyword_status', default: 'processing', null: false
```
andyduong1920: should default as `pending` instead of `processing` 🤔 it will change to processing when we actual start the process 🤔

spec/services/google/parse_service_spec.rb
andyduong1920: @Goose97 for the non-ads, count the total link or the HTML page, no need to include in any context as it's the general case

```
describe '#call' do
  it 'retrieves non-ads entries',

  it 'extracts the non-ads entry URLs',

  it 'counts the number of links',

  it 'includes the html result

  context 'given a page with top and bottom ads' do
    ...
end
```

spec/services/google/scrape_service_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+        end.to raise_error(ActiveRecord::RecordNotFound)
+      end
+
+      it 'logs an error' do
+        allow(Rails.logger).to receive(:error)
+
+        begin
+          described_class.new(keyword_id: 42).call
+        rescue ActiveRecord::RecordNotFound
+          expect(Rails.logger).to have_received(:error).with(/keyword doesn't exist/)
```
andyduong1920, line 94: In general it's ok to write more log information but for some simple case like RecordNotFound here, I think we do not need add more content for that because the default Rails framework message pretty clear already

app/commands/scrape_command.rb
Goose97: Fixed in b0713248018bbc87ba507fb4a5ffa388a807a3df

spec/services/google/scrape_service_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+RSpec.describe Google::ScrapeService, type: :service do
+  describe '#call' do
+    context 'given a VALID keyword_id' do
+      context 'given NO errors' do
+        it 'saves the html of the result page', vcr: 'google/google_no_ads' do
+          keyword = Fabricate :keyword
+
+          described_class.new(keyword_id: keyword.id).call
+          updated = Keyword.includes(:keyword_search_entries).find(keyword.id)
```
andyduong1920:
```suggestion
          updated_keyword = keyword.reload
```

spec/services/google/search_service_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+          allow(described_class).to receive(:http_client).and_return(connection)
+          adapter.get('/search') { [200, { 'Content-Type': 'text/html' }, html] }
+        end
+
+        expect(described_class.search!('hello world')).to eq(html)
+      end
+    end
+
+    context 'given a failed request' do
+      it 'raise GoogleScraperRuby::Errors::SearchServiceError if the request returns a 4xx status code' do
```
andyduong1920: `it raises`

spec/services/google/search_service_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        HttpClientHelpers.stub_http_adapter do |adapter, connection|
+          allow(described_class).to receive(:http_client).and_return(connection)
+          adapter.get('/search') { [200, { 'Content-Type': 'text/html' }, html] }
+        end
+
+        expect(described_class.search!('hello world')).to eq(html)
+      end
+    end
+
+    context 'given a failed request' do
```
andyduong1920:
```
context 'given a 2xx response`
  it 'xxx

context 'given a 4xx response`
  it 'raises xxx

context 'given a 5xx response`
  it 'raises xxx

context 'given a connection fail issue`
  it 'raises xxx
```

app/models/keyword.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Keyword < ApplicationRecord
+  enum status: { processing: 'processing', succeeded: 'succeeded', failed: 'failed' }
+
+  has_many :keyword_search_entries, dependent: :destroy
+
+  validates :content, :status, presence: true
```
andyduong1920, line 8: should we add the validation for the `result_page_html` if the status is `succeeded` @Goose97 

```
validates :result_page_html, presence: true, if: succeeded?
```

app/lib/google_scraper_ruby/errors.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module GoogleScraperRuby
+  module Errors
+    class SearchServiceError < StandardError
+      def initialize(url:, error: nil)
+        @original_error = error
+        @url = url
+        super('')
```
andyduong1920: Can we just send the original error to the parent 
```suggestion
        super(error)
```

app/services/google/search_service.rb
Goose97: Fixed in 9a60d37461cdc3d701414195e55e9592dd5e84b3

app/services/google/search_service.rb
Goose97: Fixed in 9a60d37461cdc3d701414195e55e9592dd5e84b3

app/lib/google_scraper_ruby/errors.rb
Goose97: my bad. It was a rebase error, already fixed.

.rubocop.yml
Goose97: This change was reverted based on [Oliver suggestions](https://github.com/Goose97/google-scraper-ruby/pull/20#discussion_r1280060741). Should've resolve this one, my bad.

app/services/google/scrape_service.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/b0713248018bbc87ba507fb4a5ffa388a807a3df#diff-c3acbe2de6be98fe66d47a5ec8d1965cf566a1e37807230b81b87e88c875cb27L14-L15

app/services/google/scrape_service.rb
Goose97: I intended to handle to status update part in the [background job integration ticket](https://github.com/Goose97/google-scraper-ruby/issues/15)

spec/services/google/parse_service_spec.rb
Goose97: Fixed in 99d2ec2c8842dcd2391156ebc265a4c65d733c6e

spec/services/google/scrape_service_spec.rb
Goose97: Fixed in 83904a70dcb01dce9120c1bd4bceb0109c3bb2c7

spec/services/google/scrape_service_spec.rb
Goose97: I merged two errors into one class `GoogleScraperRuby::Errors::ScrapeServiceError`. Even though it has a `kind` field to distinguish between two variants, it's not as clear as with `RecordNotFound`. So, should we keep the log?

spec/services/google/search_service_spec.rb
Goose97: Fixed in 6abe0de029aea0819556cb8a42fc90354de264fa

spec/services/google/search_service_spec.rb
Goose97: Fixed in 6abe0de029aea0819556cb8a42fc90354de264fa

app/services/google/scrape_service.rb
Goose97: Fixed in latest version

app/services/google/scrape_service.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/1ef3fbf1560eb98c39c5780571d2df1442e949a2

app/services/google/parse_service.rb
Goose97: That's a good point. I admit that there's a lot of overlapping between `KeywordSearchEntry` model and `SearchEntryStruct`. But the `KeywordSearchEntry` comes with all the baggage of `created_at`, `updated_at`, association fields, etc. I'd rather to keep these fields out of the result.

So I guess I'd keep the structs as is and add the Struct suffix as your suggestion

Goose97: Fixed in 57a134fe4ae0c6e85d29c63a25ea85e3a6ef2514

app/services/google/parse_service.rb
Goose97: Good call. Fixed in fe37c32bc8aa7ddc94fafbd011e07db7633f3d07

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+module Google
+  class ScrapeService
+    def initialize(keyword_id:)
+      @keyword_id = keyword_id
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def call!(search_service: nil)
```
olivierobert: [As commented on Slack](https://nimble-co.slack.com/archives/C05HAMT941Z/p1690897046668539), `search_service` should be added as a param in the constructor. Also, we can likely default to `Google::SearchService` if no param is provided:

```ruby
def initialize(keyword_id:, search_service: Google::SearchService)
```

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,72 @@
...
+      keyword = Keyword.find_by id: keyword_id
+      raise_keyword_not_found unless keyword
+      @keyword = keyword
+
+      begin
+        search_service ||= Google::SearchService.new(keyword.content)
+        html = search_service.search!
+      rescue GoogleScraperRuby::Errors::SearchServiceError => e
```
olivierobert: Avoid [abbreviated variable names](https://nimblehq.co/compass/development/code-conventions/#convention-12):

```suggestion
      rescue GoogleScraperRuby::Errors::SearchServiceError => error
```

Prefer spelling out every variable name. You are not saving many keystrokes 😉  and it helps readability in the long run.

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,72 @@
...
+    def save_scrape_result(keyword, parse_result)
+      Keyword.transaction do
+        keyword.update!(
+          links_count: parse_result.links_count,
+          result_page_html: parse_result.result_page_html,
+          status: :succeeded
+        )
+
+        # rubocop:disable Rails/SkipsModelValidations
+        keyword.keyword_search_entries.insert_all parse_result.search_entries.map(&:to_h)
```
olivierobert: As raised before, it would be better to be consistent in the usage of parenthesis. Either use them all them time or never, but not only sometimes. In this case, parenthesis is used for `map` but not `insert_all` 🙃 

```suggestion
        keyword.keyword_search_entries.insert_all(parse_result.search_entries.map(&:to_h))
```

app/services/google/search_service.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def initialize(keyword)
+      @keyword = keyword
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def search!
+      uri = URI::HTTPS.build host: HOST, path: '/search', query: { q: keyword }.to_query
+      conn = http_client uri
+      response = conn.get uri, {}
```
olivierobert: Let's add a line break here too:

```suggestion
      response = conn.get uri, {}
      
```

app/services/google/search_service.rb
```diff
@@ -0,0 +1,59 @@
...
+    private_constant :HOST, :USER_AGENT
+
+    def initialize(keyword)
+      @keyword = keyword
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def search!
+      uri = URI::HTTPS.build host: HOST, path: '/search', query: { q: keyword }.to_query
+      conn = http_client uri
```
olivierobert: It seems like you missed my previous comment about not using the abbreviated name coming from the Phoenix world 😅 

```suggestion
      client = http_client(uri)
```

olivierobert (commented):
The tests part is much better now, but make sure not to miss any previously raised comments, to avoid lengthy back and forth.

app/services/google/search_service.rb
Goose97: Fixed in 9b3698824c9292c74247204b22f7df5c0fbed19c

app/models/keyword_search_entry.rb
Goose97: Somehow, I can not find a search keyword that will show these kind of ads so I can create a vcr fixture for it.

What is your search phrase?

app/models/keyword.rb
Goose97: Nice catch. Fixed in 77490778458b6389e4af1cc565d529aa829ee7cc

app/services/google/scrape_service.rb
Goose97: Fixed in 6fbb5e0caa34d9cb91ab35a3f314d40cac37eb95

app/services/google/scrape_service.rb
Goose97: Fixed in 19712c4f4945fdd0cb24eb04d37db0d532b43fa5

app/lib/google_scraper_ruby/errors.rb
```diff
@@ -0,0 +1,46 @@
...
+          - original_error: #{original_error}"
+        ERR_MESSAGE
+      end
+
+      private
+
+      attr_reader :original_error, :url
+    end
+
+    class ScrapeServiceError < StandardError
```
andyduong1920: in Ruby we would prefer to have separate file for each class instead of having a single file for all classes @Goose97 so let's move this to 2 files 

- app/lib/google_scraper_ruby/errors/search_service_error.rb
- app/lib/google_scraper_ruby/errors/scrape_service_error.rb

I also recommend to remove the `_service_` in the file name, so it's just about the `search_error` or `scrape_error` what do you think?

app/models/keyword.rb
```diff
@@ -0,0 +1,20 @@
...
+  validates :result_page_html, presence: true, if: :succeeded?
+  validates :links_count,
+            presence: true,
+            numericality: { greater_than_or_equal_to: 0 },
+            if: :succeeded?
+
+  private
+
+  def succeeded?
```
andyduong1920: with [Enum](https://api.rubyonrails.org/v7.0.6/classes/ActiveRecord/Enum.html), it will define the `succeeded?` method for us  FYI, so we can remove this one ;)

app/models/keyword_search_entry.rb
andyduong1920: my keyword there is just `Steam deck` @Goose97 :D

app/services/google/scrape_service.rb
andyduong1920: Shall we add the TODO so the reviewer will aware about your plan @Goose97

spec/services/google/scrape_service_spec.rb
andyduong1920: In most case with the `RecordNotFound` we know that the Keyword doesn't exist :D but you can keep the log there 🙆‍♂️

app/services/google/parse_service.rb
andyduong1920: > But the KeywordSearchEntry comes with all the baggage of created_at, updated_at, association fields, etc.

Is there any issue with that @Goose97 for us, we don't see any issue, also the created_at, updated_at doesn't have any value when we `new` the KeywordSearchEntry object FYI, I have tested with a model on my local

image

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,70 @@
...
+module Google
+  class ScrapeService
+    def initialize(keyword_id:, search_service: Google::SearchService.new)
+      @keyword_id = keyword_id
+      @search_service = search_service
+    end
+
+    def call!
+      keyword = Keyword.find_by id: keyword_id
+      raise_keyword_not_found unless keyword
```
andyduong1920: why not just use the `Keyword.find(keyword_id)` like previous version and we catch on the RecordNotFound exception like before 🤔

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,70 @@
...
+      @keyword = keyword
+
+      begin
+        html = search_service.search! keyword.content
```
andyduong1920: how about move this into a separate method

```
html_response = search

private

def search
  search_service.search! keyword.content
rescue GoogleScraperRuby::Errors::SearchServiceError => error
        raise_unexpected_error error
end
```

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,70 @@
...
+      rescue GoogleScraperRuby::Errors::SearchServiceError => error
+        raise_unexpected_error error
+      end
+
+      result = Google::ParseService.new(html).call
```
andyduong1920: Above we pass the `search_service` and here we use the `Google::ParseService` directly, so we keep the pattern consistency, should we pass the `parse_service` to this as well 🤔

db/migrate/20230727072004_create_keywords.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    create_enum  :search_entry_kind, %w[ads non_ads]
+
+    create_enum  :search_entry_position, %w[top bottom]
+
+    create_table :keyword_search_entries, id: :serial do |t|
+      t.enum :kind, enum_type: 'search_entry_kind', null: false
+      t.string :urls, array: true, default: []
+      t.enum :position, enum_type: 'search_entry_position'
```
andyduong1920:
```suggestion
      t.enum :position, enum_type: 'search_entry_position', null: false
```

spec/services/google/parse_service_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        e.kind == :non_ads && e.urls.any? { |u| /ruby-lang.org/ =~ u }
+      end
+
+      expect(entry).not_to be_nil
+    end
+
+    it 'counts the number of links', vcr: 'google/monitor_50_links' do
+      html = Google::SearchService.new.search! 'monitor'
+
+      count = described_class.new(html).call.links_count
```
andyduong1920:
```suggestion
      links_count = described_class.new(html).call.links_count
```

or we can just test it directly

```
expect(described_class.new(html).call.links_count).to eq(50)
```

spec/services/google/parse_service_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      expect(count).to eq(50)
+    end
+
+    it 'includes the html of such page', vcr: 'google/monitor_50_links' do
```
andyduong1920:
```suggestion
    it 'includes the html of the page', vcr: 'google/monitor_50_links' do
```

spec/services/google/scrape_service_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+          expect(updated_keyword.result_page_html).to be_a String
+        end
+
+        it 'saves the links count', vcr: 'google/google_no_ads' do
+          keyword = Fabricate :keyword
+
+          described_class.new(keyword_id: keyword.id).call!
+          updated_keyword = keyword.reload
+
+          expect(updated_keyword.links_count).to be_a Integer
```
andyduong1920: should we test with the actual value, like 1 or 2?

spec/services/google/scrape_service_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+
+        it 'saves all search entires', vcr: 'google/google_no_ads' do
```
andyduong1920: should we test it updates the status to `succeed`?

app/models/keyword.rb
Goose97: Fixed in 3f79f1904b1d6154608ebca568f0920f040814ec

app/services/google/scrape_service.rb
Goose97: ah yeah, I was thinking about it as well but decide to leave it like this. No particular reason. Your point makes sense

Goose97: Fixed in b5840010db8bda70161194af39239f14661508f9

app/services/google/scrape_service.rb
Goose97: yeah, sure. Fixed in e64878013a57db1b0ec6884073940edd47c5a74e

db/migrate/20230727072004_create_keywords.rb
Goose97: Good point. Fixed in 5e498403fb9cb5f3583c9e5f701826ce34f60c42

app/services/google/parse_service.rb
Goose97: Oh, my rationale here is: I want to keep the ParseService to detach from the database layer. Fields like `created_at`, `updated_at` or associations are database concepts, whether or not they have value. This is just my preferences, no noticeable runtime differences.

However, after weighing the pros and cons, I think it better to stick with the team conventions here.

Goose97: Fixed in bb8653ab5d4283653e4512deee39fc3758aecb96

spec/services/google/scrape_service_spec.rb
Goose97: Fixed in b702425f6dac9764fb1c4a4c391774acddc991a1

olivierobert (approved):
One Andy's comments are resolved, this can be merged. So giving a pre-approval for not blocking you :-)

app/lib/google_scraper_ruby/errors.rb
Goose97: Fixed in https://github.com/Goose97/google-scraper-ruby/pull/20/commits/54cff3f50e72c4bd4d34390f93d1c25caf25b5bc

spec/services/google/scrape_service_spec.rb
Goose97: If so, should we update the fixture name to reflect that?

from
```
google/google_no_ads
```

to
```
google/google_no_ads_N_links
```

app/models/keyword_search_entry.rb
Goose97: Still no ads :(. Can we leave this one for future releases?

spec/services/google/scrape_service_spec.rb
andyduong1920: That is not necessary I think @Goose97 or you can test the `expect(updated_keyword.links_count).to be > 0`

app/lib/google_scraper_ruby/errors.rb
andyduong1920: @Goose97 the fix is just about rename the class, both class still in the same file, please double check my above comment 🙏

db/migrate/20230803022448_add_keyword_status_variant.rb
```diff
@@ -0,0 +1,9 @@
+class AddKeywordStatusVariant < ActiveRecord::Migration[7.0]
+  def change
+    execute <<-SQL
+      ALTER TYPE keyword_status ADD VALUE 'pending';
+    SQL
+
+    change_column_default :keywords, :status, 'pending'
```
andyduong1920: FYI @Goose97 as we haven't merge this migration so with this kind of new change we can also edit it directly in the original migration :D app/services/google/search_service.rb, but a new migration like this as also fine just FYI

db/schema.rb
```diff
@@ -0,0 +1,43 @@
...
+  # Custom types defined in this database.
+  # Note that some types may not work with other database engines. Be careful if changing database.
+  create_enum "keyword_status", ["processing", "succeeded", "failed", "pending"]
+  create_enum "search_entry_kind", ["ads", "non_ads"]
+  create_enum "search_entry_position", ["top", "bottom", "main_search"]
+
+  create_table "keyword_search_entries", id: :serial, force: :cascade do |t|
+    t.enum "kind", null: false, enum_type: "search_entry_kind"
+    t.string "urls", default: [], array: true
+    t.enum "position", enum_type: "search_entry_position"
```
andyduong1920: Small thing, should we add `null: false`?

spec/services/google/parse_service_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      expect(links_count).to eq(50)
+    end
+
+    it 'includes the html of the page', vcr: 'google/monitor_50_links' do
+      html = Google::SearchService.new.search! 'monitor'
+
+      result_page = described_class.new.call(html).result_page_html
+
+      expect(result_page).to be_a String
```
andyduong1920: should we test the String here is not an empty string `''` 😄

app/lib/google_scraper_ruby/errors.rb
Goose97: my bad. Fixed in 8d8f1ce6d19a1481c0ac1e8b21c7c935ce4f4bd0

spec/services/google/parse_service_spec.rb
Goose97: Fixed in 9db72f6cbabd519c0a1d76cf3c1a887383cf9bf2

db/schema.rb
Goose97: Fixed in 7a92f96ad741fbd954383290d8ccfde4d7326772

Also merge migration files to keep it tidy

db/migrate/20230803022448_add_keyword_status_variant.rb
Goose97: Thanks, good to know

spec/services/google/scrape_service_spec.rb
Goose97: Fixed in 511e334d416ff1d2b41e9419cbd7f479575f16f2

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,67 @@
...
+      ERROR
+
+      raise GoogleScraperRuby::Errors::ScrapeError.new(
+        keyword_id: keyword_id,
+        kind: :unexpected_error,
+        error: error
+      )
+    end
+
+    def save_scrape_result(keyword, parse_result)
```
andyduong1920:
```suggestion
    def save_scrape_result!(keyword, parse_result)
```

db/migrate/20230727072004_create_keywords.rb
andyduong1920:
```suggestion
      t.enum :status, enum_type: 'keyword_status', default: 'pending', null: false
```

app/services/google/scrape_service.rb
Goose97: oh I think this makes it clearer where the error came from. Try to avoid potential confusion

app/services/google/scrape_service.rb
Goose97: Sure 22323d5506c9b7c5b6d0f0ff979c9b97fac3df4b

app/services/google/scrape_service.rb
andyduong1920: no need to do that @Goose97 as in Rails, we know that the `find(xx)` will raise the NotFound exception, so we just catch it.

db/migrate/20230727072004_create_keywords.rb
Goose97: Oops, git actions gone rogue here. Fixed in 91c28f7a80275a762abfc0bb70c22db03b15c423

app/services/google/scrape_service.rb
Goose97: Fixed in da5fc27e929309dee83eb89e49dc76f983eb393b

app/services/google/scrape_service.rb
Goose97: Thanks for the note. Fixed in 736b1d120535fc0713ebd5fa47eb230a634031b8

app/services/google/scrape_service.rb
```diff
@@ -0,0 +1,68 @@
...
+  class ScrapeService
+    def initialize(keyword_id:, search_service: Google::SearchService.new, parse_service: Google::ParseService.new)
+      @keyword_id = keyword_id
+      @search_service = search_service
+      @parse_service = parse_service
+    end
+
+    def call!
+      # TODO(Goose97): update failed status once we integrate with background job
+      begin
```
andyduong1920, line 13: just a small thing, we do not need the `begin` here @Goose97 so you can delete it

andyduong1920 (approved):
Just a small comment, once it fixed we can merge this PR @Goose97 💪

Issue #9, opened 2023-07-24
[Backend] As a User, I can upload CSV keyword files


Goose97:
Backend for #8 

## Why

Authenticated users should be able to provide keywords to scrape. To achieve that, we provide an interface for users to upload CSV files.

## Acceptance Criteria

- Authenticated users must be able to upload CSV files containing keywords
- Keyword files should be validated:
  - A keyword file must contains between 1 and 1000 keywords
  - A keyword file must have size limit of 5MB
- After upload, all keywords in the file should be persisted in the database
- Do NOT process the keyword file. This will tackled in #10

olivierobert:
The note should be part of the ACs (also tests must be added by default so they can be skipped):

```
- Authenticated users must be able to upload CSV files containing keywords
- Validate the keyword file:
   - A keyword file must contains between 1 and 1000 keywords
   - A keyword file must have size limit of 5MB
- After upload, all keywords in the file should be persisted in the - 
- Do NOT process the keyword file. This will tackled in #10.
```

ACs can contain both what to do and what NOT to do (see the emphasis on the NOT for readability).

PR #30, opened 2023-08-01
[Backend] As a User, I can upload CSV keyword files

Goose97:
[[Backend] As a User, I can upload CSV keyword files](https://github.com/Goose97/google-scraper-ruby/issues/9)

## What happened 👀

Implement backend side for CSV upload feature

## Insight 📝

- Implement CsvUploadForm: this class encapsulates all validation and database interaction logics


olivierobert (commented):
You need to branch out from `feature/scrape-keyword` so that only the changes for the cs upload are reviewable 🙈

Goose97:
> You need to branch out from `feature/scrape-keyword` so that only the changes for the cs upload are reviewable 🙈

Oops, my bad 🥶 . I created a new PR with the correct branch name here https://github.com/Goose97/google-scraper-ruby/pull/31

PR #31, opened 2023-08-02
[Backend] As a User, I can upload CSV keyword files

### LINKED ISSUE #9: [Backend] As a User, I can upload CSV keyword files

Goose97:
Closes [[Backend] As a User, I can upload CSV keyword files](https://github.com/Goose97/google-scraper-ruby/issues/9)

## What happened 👀

Implement backend side for CSV upload feature

## Insight 📝

- Implement CsvUploadForm: this class encapsulates all validation and database interaction logics


Goose97:
This PR include some model files from this PR https://github.com/Goose97/google-scraper-ruby/pull/20 to make the tests work. Once https://github.com/Goose97/google-scraper-ruby/pull/20 is merged, this branch will be rebased to erase those files

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  attr_reader :file
+
+  def parsed_keywords
+    CSV.read(file).filter_map do |record|
+      content = record.join ','
+      unless content.strip.empty?
+        {
+          content: content,
+          status: :processing
```
andyduong1920: do we need to set the status here? or should we let the default value from the db migration handle that?

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  validates :file, presence: true, valid_keyword_file: true
+
+  def save(file)
+    @file = file
+    return false unless valid?
+
+    begin
+      # rubocop:disable Rails/SkipsModelValidations
+      @keyword_ids = Keyword.insert_all(parsed_keywords).pluck('id')
```
andyduong1920: Just curious how will you plan to start the search for each keyword in the next PR, will you have a cronjob to check the new added record or how, because with the `insert_all` there will be no callback event like `after_commit` in the model https://guides.rubyonrails.org/active_record_callbacks.html#skipping-callbacks

app/validators/valid_keyword_file_validator.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+class ValidKeywordFileValidator < ActiveModel::EachValidator
```
andyduong1920: How about 
```suggestion
class CSVKeywordFileValidator < ActiveModel::EachValidator
```

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+class CsvUploadForm
+  include ActiveModel::Validations
```
andyduong1920: it's ok to include just the ActiveModel::Validations here but FYI we could also includes all other ActiveRecord helpers to this Form with `include ActiveModel::Model`

spec/forms/csv_upload_form_spec.rb
```diff
@@ -0,0 +1,122 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'support/file_upload_helpers'
+
+RSpec.describe CsvUploadForm, type: :form do
+  describe '#save' do
+    context 'given a VALID csv file' do
```
andyduong1920:
```suggestion
    context 'given a VALID CSV file' do
```

spec/forms/csv_upload_form_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+
+      context 'given a file with incorrect content type' do
+        it 'adds an error' do
+          form = described_class.new
+          file = FileUploadHelpers.upload_file fixture: 'wrong_type.txt', content_type: 'text/plain'
+
+          form.save file
+
+          expect(form.errors.messages).to include(
+            file: include(I18n.t('csv.errors.invalid_keyword_count'))
```
andyduong1920: is it `invalid_keyword_count` message 🤔

spec/forms/csv_upload_form_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+      end
+
+      context 'given some blank keywords' do
+        it 'skips those keywords' do
+          form = described_class.new
+          file = FileUploadHelpers.upload_file fixture: 'blank_keywords.csv'
+
+          expect do
+            form.save file
+          end.to change(Keyword, :count).by(6)
```
andyduong1920: this test doesn't say much, we don't know what is the 6 keywords there as the file is `blank_keywords.csv` 🤔 

how about changing the fixture file to `6_keywords_and_blank_keywords.csv`

spec/forms/csv_upload_form_spec.rb
Goose97: good catch. Totally missed that one.

spec/forms/csv_upload_form_spec.rb
Goose97: Good call. Fixed in 760fe8fa9cef05273db5a2562331cbf66ddd4d3c

spec/forms/csv_upload_form_spec.rb
Goose97: Fixed in ab157a4212523e8ffed59a8096fa4e5377ae8a04

app/forms/csv_upload_form.rb
Goose97: yeah, my bad. Missed that line when reading the compass. adf505e4a0180d1abfdc0defb34ad7c441089a33

app/controllers/keywords_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class KeywordsController < ApplicationController
+  def create
+    form = CsvUploadForm.new
+
+    if form.save create_params[:file]
```
olivierobert: Again raising the concern of not using parenthesis consistently. Since you use them with I18n.t, this should be:

```suggestion
    if form.save(create_params[:file])
```

I recommend you to stick to using parenthesis to avoid any inconsistency issues 🤞

spec/requests/keywords_controller_spec.rb
olivierobert, line 1: IT is best to prefer request specs. See https://nimblehq.co/compass/development/code-conventions/ruby/rspec/#convention-13

spec/fixtures/files/too_long_keywords.csv
```diff
@@ -0,0 +1,4 @@
+Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.
+the rest is discarded
+
+
```
olivierobert: Are the two extra line breaks needed/intended? If not, they should be removed.

config/locales/en.yml
```diff
@@ -31,3 +31,9 @@
 
 en:
   hello: "Hello world"
```
olivierobert: Lets remove the boilerplate code 🙈

app/validators/valid_keyword_file_validator.rb
Goose97: Fixed in 315cbbc32f5c7012278e5a65246259d9948b0a26

config/locales/en.yml
```diff
@@ -31,3 +31,9 @@
...
+  csv:
```
olivierobert: These locales should be in a file called `activemodel.en.yml`. See: https://github.com/nimblehq/internal-certification-ruby/blob/main/huey-toby/config/locales/activemodel.en.yml

Rails can automatically load translations based on file names. See: https://guides.rubyonrails.org/i18n.html#lazy-lookup

olivierobert (changes requested):

app/forms/csv_upload_form.rb
Goose97: Fixed in 1c83b31647d79b6a3058881ecec0ace5d92727bc

app/forms/csv_upload_form.rb
Goose97: yeah I intend to ask you about this. 

I'm thinking about using create
```
# from
@keyword_ids = Keyword.insert_all(parsed_keywords).pluck('id')

# to
@keyword_ids = Keyword.create(parsed_keywords).pluck('id')
```

I see that it performs a SQL query per item. Should I worry about it? Not that I'll affect this project much, just wonder about the general way to approach this.

app/controllers/keywords_controller.rb
Goose97: Remove parentheses of I18n eb74e685b4c239c4ccfafb68461b3be159e4a296

spec/requests/keywords_controller_spec.rb
Goose97: Just to be clear. Is this just about changing the file name, right? AFAIK, the test cases already follow [request specs guidelines](https://rspec.info/features/6-0/rspec-rails/request-specs/request-spec/)

Fixed in 7ca000c6c73670af0feb7dab2a25a80083add812

spec/fixtures/files/too_long_keywords.csv
Goose97: Fixed in bb5336c6dad9bf5bf54e9af7652654b36a3b3cbe

config/locales/en.yml
Goose97: Fixed in 366f0e361c9209f473a876ffbab0e837add6eb33

config/locales/en.yml
Goose97: Fixed in 366f0e361c9209f473a876ffbab0e837add6eb33

app/controllers/keywords_controller.rb
Goose97: I thinks it's too late to go back to the no parens style now 😥 . Just to be clear, as long as we stick to one style and be consistent, we should be ok, right?

app/forms/csv_upload_form.rb
andyduong1920: Regarding the performance, we shouldn't worry it for now, just let's just insert 1 by 1 keyword for now, so with that you can use the ActiveRecord callbacks (after_commit, after_create) to trigger the background job later

app/controllers/keywords_controller.rb
andyduong1920: @Goose97 in Ruby we prefer the parenthesis to reduce the confusion, the lining tool could help you to add all nessesarry parenthesis automatically, you can do it by

1. Update the `rubocop.yml`
```
Style/MethodCallWithArgsParentheses:
  Enabled: true
```
2. Run `rubocop -a`

More information you can check on https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MethodCallWithArgsParentheses

spec/forms/csv_upload_form_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+      it 'returns true' do
+        form = described_class.new
+        file = FileUploadHelpers.upload_file fixture: 'valid_7_keywords.csv'
+
+        success = form.save file
+
+        expect(success).to be(true)
+      end
+
+      it 'save all keywords' do
```
andyduong1920:
```suggestion
      it 'saves all keywords' do
```

app/controllers/keywords_controller.rb
andyduong1920: I also created an RFC for that https://github.com/nimblehq/compass/discussions/1315 so if the team vote for Yes, we can enforce it by modify the rubocop.yml in the Rails template ;)

app/forms/csv_upload_form.rb
Goose97: great, got it. Fixed in 9162b42eee632a76a818c3ded71d0bb68ccee90a

spec/forms/csv_upload_form_spec.rb
Goose97: Fixed in 73e3ef124168274de87342bde97fadb897dd54db

app/controllers/keywords_controller.rb
Goose97: yeah, I prefer the parentheses as well. It's also the Elixir's choice regarding this matter.

I thought that Ruby prefers the parentheses-less style (don't know where I got this idea 😅) and stuck with it from the beginning

andyduong1920: @Goose97 let's force the parenthesis by update the rubocop.yml file and run the `rubocop -a` on this PR 👌

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  attr_reader :keyword_ids
+
+  validates :file, presence: true, csv_keyword_file: true
+
+  def save(file)
+    @file = file
+    return false unless valid?
+
+    begin
```
andyduong1920: Just a small thing, we can omit the `begin` here, that is not a big topic @Goose97 :)

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,35 @@
...
+      @keyword_ids = Keyword.create(parsed_keywords).pluck('id')
+      true
+    rescue ActiveRecord::ValueTooLong
+      errors.add(:file, I18n.t('activemodel.csv.errors.invalid_keyword_length'))
```
andyduong1920: I'm curious, let's say this error is happening on the middle of the process, will all created keywords in this process get rollback? Should we add a test for that @Goose97 

```
# in the same file
1. valid_keyword_1 -> created
2. valid_keyword_2 -> created
3. exceed_255_keyword -> error 💥 
4. ...
```

will the **valid_keyword_1 and valid_keyword_2** got deleted?

app/validators/csv_keyword_file_validator.rb
```diff
@@ -0,0 +1,24 @@
...
+    record.errors.add(attribute, I18n.t('activemodel.csv.errors.invalid_keyword_count')) unless valid_record_count?
+    record.errors.add(attribute, I18n.t('activemodel.csv.errors.invalid_file_type')) unless valid_content_type?
+  end
+
+  private
+
+  attr_reader :file
+
+  def valid_record_count?
+    CSV.read(file).count.between?(1, 1000)
```
andyduong1920: should we filter the `empty keyword` with `content.strip.empty?` 🤔

app/validators/csv_keyword_file_validator.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+class CsvKeywordFileValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    @file = value
+
+    record.errors.add(attribute, I18n.t('activemodel.csv.errors.invalid_keyword_count')) unless valid_record_count?
+    record.errors.add(attribute, I18n.t('activemodel.csv.errors.invalid_file_type')) unless valid_content_type?
```
andyduong1920, line 6: should we check the type before the `record_count`?

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,35 @@
...
+  validates :file, presence: true, csv_keyword_file: true
+
+  def save(file)
+    @file = file
+    return false unless valid?
+
+    begin
+      @keyword_ids = Keyword.create(parsed_keywords).pluck('id')
+      true
+    rescue ActiveRecord::ValueTooLong
```
andyduong1920: btw should we move this to inside the `validator` like `keywords.any? { |keyword| keyword.length > 255 } { I18n.t('invalid_keyword_length') } so all validation will be just in that validator

spec/forms/csv_upload_form_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+
+          form.save file
+
+          expect(form.errors.messages).to include(
+            file: include(I18n.t('activemodel.csv.errors.invalid_keyword_count'))
+          )
+        end
+      end
+
+      context 'given a too long keyword' do
```
andyduong1920:
```suggestion
      context 'given the keyword that has more than 255 characters' do
```

app/controllers/keywords_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      flash[:notice] = I18n.t 'csv.upload_success'
+    else
+      flash[:alert] = form.errors.full_messages
+    end
+
+    redirect_to keywords_path
```
andyduong1920: Should we define the `keywords#index` action now? otherwise I'm not sure where does the `keywords_path` come from 🤔

Goose97: Fixed in 101795de4440249c337c9f901659c0c01accd3c4

app/validators/csv_keyword_file_validator.rb
Goose97: oh good one, did not notice that. c20f36211bb1f0c2b72d8a8862b690f328dcea1a

app/forms/csv_upload_form.rb
Goose97: Fixed in 0c4e8d7914b6fc71ddb701a3d3003cc1c1730124

app/validators/csv_keyword_file_validator.rb
Goose97: Fixed in 0c4e8d7914b6fc71ddb701a3d3003cc1c1730124

app/forms/csv_upload_form.rb
Goose97: With the validation logic now all moved into the Validator class, we can omit the rescue block. In this case, I think we can safely ignore this scenario.

What do you think?

app/forms/csv_upload_form.rb
Goose97: Removed the rescue block here cd0402d71dd82d0fbabe17c4271371a605df2f42

app/controllers/keywords_controller.rb
Goose97: One small thing. Does this rule apply to spec files as well?

Goose97: Fixed in 8aca492dbd8f4380a2e7f1c58c429fad1088359c

andyduong1920: > Does this rule apply to spec files as well?

@Goose97 let apply it there as well 👍

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+class CsvUploadForm
+  include ActiveModel::Model
+
+  attr_reader :keyword_ids
+  attr_accessor :keywords
```
andyduong1920: Not sure why do we need the `  attr_accessor :keywords` here

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  validates :file, presence: true, csv_keyword_file: true
+
+  def save(file)
+    @file = file
+    return false unless valid?
+
+    entries = keywords.map { |keyword| { content: keyword } }
+    @keyword_ids = Keyword.create(entries).pluck('id')
```
andyduong1920: I think we don't need to to use `@keyword_ids` here, with the `Keyword.create` we could use the Active Record callback to trigger the Background job 1 by 1

app/validators/csv_keyword_file_validator.rb
```diff
@@ -0,0 +1,38 @@
...
+
+require 'csv'
+
+class CsvKeywordFileValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    @file = value
+    @keywords = parsed_keywords
+    record.keywords = @keywords
+
+    record.errors.add(attribute, I18n.t('activemodel.csv.errors.invalid_file_type')) unless valid_content_type?
```
andyduong1920, line 6: should we do this validate before the parsed_keyword step?

app/validators/csv_keyword_file_validator.rb
```diff
@@ -0,0 +1,38 @@
...
+      content unless content.empty?
+    end
+  end
+
+  def valid_record_count?
+    keywords.length.between?(1, 1000)
+  end
+
+  def valid_keyword_length?
+    keywords.all? { |keyword| keyword.length <= 255 }
```
andyduong1920: instead loop over all keyword, just use `any?`

```
def invalid_keyword_length
  keywords.any? { |keyword| keyword.length > 255 }
end
```

app/forms/csv_upload_form.rb
Goose97: I want the validator to parse the keywords and returns it to the form object (by setting it the `keywords` field) https://github.com/Goose97/google-scraper-ruby/blob/8aca492dbd8f4380a2e7f1c58c429fad1088359c/app/validators/csv_keyword_file_validator.rb#L8

app/validators/csv_keyword_file_validator.rb
Goose97: Good call. Fixed in 360a4a0621b88eb9bbc475bade0d53fb464237f7

app/validators/csv_keyword_file_validator.rb
Goose97: I think both cases have identical behavior, aren't they? Is this just the inverse of method name?

app/forms/csv_upload_form.rb
Goose97: Fixed in 63d275321ec7666c28d8d0bdc6ea4a94f61645f2

app/forms/csv_upload_form.rb
Goose97: We still need to test that keywords file doesn't get processed partially. Here's the test case for that 22fb4488dab5685eb7201d9fac14743512bc57c3

app/forms/csv_upload_form.rb
andyduong1920: Oh I see, but can we try with the `attr_reader` only @Goose97 🤔

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,24 @@
...
+  attr_accessor :keywords
+
+  validates :file, presence: true, csv_keyword_file: true
+
+  def save(file)
+    @file = file
+    return false unless valid?
+
+    entries = keywords.map { |keyword| { content: keyword } }
+    Keyword.create(entries).pluck('id')
```
andyduong1920:
```suggestion
    Keyword.create(entries)
```

andyduong1920: If one of the keyword failed to be created, will the whole transaction get rollback? Also after checking the `create with entries` also `inserted one-at-a-time into the database.` it mean it will be hard for you to trigger the background job later, am I right?

app/validators/csv_keyword_file_validator.rb
andyduong1920: the `any` will exist if the condition match, so for example we have 200 keywords and the 5th keyword is invalid, the `keywords.any? { |keyword| keyword.length > 255 }` will exist, no need to check for the rest https://www.rubyguides.com/2018/10/any-all-none-one/

app/forms/csv_upload_form.rb
andyduong1920: for me an easy case for now is just creating multiple keyword, that is an easier solution

```
ActiveRecord::Base.transaction do
  keywords.each do |keyword|
    Keyword.create(content: keyword)
  end
end
```

spec/requests/keywords_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+    end
+
+    context 'given an INVALID file' do
+      it 'includes a flash message in the response' do
+        file = FileUploadHelpers::Form.upload_file(fixture: 'too_many_keywords.csv')
+
+        post keywords_path, params: { csv_upload_form: { file: file } }
+
+        expect(flash[:alert]).not_to(be_nil)
```
andyduong1920:
```suggestion
        expect(flash[:alert]).to(eq(I18n.t('activemodel.csv.errors.invalid_keyword_count')))
```

andyduong1920 (commented):
Just some small comments @Goose97

app/controllers/keywords_controller.rb
Goose97: all done 💪

app/forms/csv_upload_form.rb
Goose97: yeah, I've tried it, but it doesn't seem to work

```
  1) CsvUploadForm#save given a VALID CSV file returns true
     Failure/Error: record.keywords = @keywords
     
     NoMethodError:
       undefined method `keywords=' for #<CsvUploadForm:0x0000000107f4b5c0 @file=#<Rack::Test::UploadedFile:
0x0000000107f4a300 @content_type="text/csv", @original_filename="valid_7_keywords.csv", @tempfile=#<Tempfile
:/var/folders/l_/qdkzttfx02sg4_h338dzdy500000gn/T/valid_7_keywords20230808-9125-sep665.csv>>, @validation_co
ntext=nil, @errors=#<ActiveModel::Errors []>>
     # ./app/validators/csv_keyword_file_validator.rb:11:in `validate_each'
     # ./app/forms/csv_upload_form.rb:14:in `save'
     # ./spec/forms/csv_upload_form_spec.rb:13:in `block (4 levels) in <top (required)>'
     # ./spec/support/vcr.rb:31:in `block (2 levels) in <top (required)>'
```

app/validators/csv_keyword_file_validator.rb
Goose97: Thanks for the explanation. But somehow, I still don't get it 😿 

I think your explanation also applies to `all?`, right? If the 5th keyword is invalid, the function will return false and the rest are skip

```ruby
list = [1, 2, 3]
list.all? do |x|
  puts x
  x < 2
end

# It prints
1
2
```

app/forms/csv_upload_form.rb
Goose97: My bad. I was suspicious about whether Keyword.create uses transactions or not. I inspected the log from the test output and saw a transaction block. But turns out, the one who performed the transaction is Rspec, not ActiveModel as I assumed 🥶 
Screenshot 2023-08-08 at 10 32 03

Anw, 333e81a2617b5688c2afa2b0b65a4ac6fe04f200 fixed this

app/services/google/scrape_service.rb
```diff
@@ -11,50 +11,56 @@ def initialize(keyword_id:, search_service: Google::SearchService.new, parse_ser
...
     private
 
     attr_reader :keyword_id, :keyword, :search_service, :parse_service
 
     def result_page!
-      search_service.search! keyword.content
+      search_service.search!(keyword.content)
     rescue GoogleScraperRuby::Errors::SearchError => error
-      raise_unexpected_error error
+      raise_unexpected_error(error)
```
olivierobert: Technical every caught error is not unexpected 🙈 Here is can be:

```suggestion
      raise_srape_error(error)
```

It matches the custom error class name too.

app/forms/csv_upload_form.rb
olivierobert: That introduces a very tight coupling between two objects, which is not ideal. It is really unexpected to rely on a validator class to do the CSV parsing logic, breaking the single responsibility principle. 

The code architecture should be:
- Form object process the files, hence does the parsing
- The validator should only do the file validation, i.e., check-in file type, check if not empty)
- Implement other validations for keyword count and length

If you prefer to decouple things even further, the CSV parsing can be handled by another object, but it kind of seems over-engineering unless we need to handle many file types. In that case, we would need multiple parsers, and we could use Dependency Injection to set which parser to use in the model.

olivierobert (changes requested):

app/services/google/scrape_service.rb
Goose97: Fixed in aebdfd3c7bbb57207b42bf3b24ca2788ca6886d0

app/forms/csv_upload_form.rb
Goose97: Good call on the single responsibility principle. The `keywords` field is the by-product of me moving to many logic into the validator class in the first place.

Fixed in 8e13d65a21917cd5f8e9b905627e1c234d59cb13

olivierobert (approved):
The custom validator is now a bit empty 😅 but let's move forward with this implementation ✌️

app/validators/csv_keyword_file_validator.rb
andyduong1920: oh the `all` also break on the first fail condition, good to know that @Goose97 👍

andyduong1920 (approved):

PR #32, opened 2023-08-07
[Backend] As a User, my uploaded keywords should be processed immediately

Goose97:
[[Backend] As a User, my uploaded keywords should be processed immediately](https://github.com/Goose97/google-scraper-ruby/issues/15)

## What happened 👀

Integrate keyword scraping with background job system backed by Sidekiq

## Insight 📝

- Jobs are enqueued immediately after keywords are created via `after_create_commit` callback
- The retry mechanism is currently implemented using ActiveJob `retry_on` method. Another alternative that I've considered is using Sidekiq native retry mechanism. I chose ActiveJob because:
  - After looked up others IC, it seems like we prefer using ActiveJob
  - Handling retry times exhaustion (in this case, we set keyword status to `:failed`) with ActiveJob is more convenient. Sidekiq has an option to do that, but ActiveJob is [yet to support it](https://github.com/sidekiq/sidekiq/pull/5994)
  
~~There is one caveat when choosing ActiveJob. My hunch is that if the Ruby node is restarted when a job is processing/retrying, the job will be marked as failed and since [we disable Sidekiq retry](https://github.com/Goose97/google-scraper-ruby/blob/6ea9f10725679d8fa43749ab837bae6754cbee43/app/jobs/scrape_keyword_job.rb#L6), the job will be lost. I'm looking for documentations to confirm this. Can I have your words on this @andyduong1920 ?~~

My bad, I thought that background worker threads are run inside Rails server process. They run in Sidekiq process, so restart the server process won't affect running jobs

### DIFF
--- a/app/jobs/scrape_keyword_job.rb
+++ b/app/jobs/scrape_keyword_job.rb
@@ -16,6 +16,6 @@ class ScrapeKeywordJob < ApplicationJob
   def perform(keyword_id:)
     Google::ScrapeService.new(keyword_id: keyword_id).call!
   rescue GoogleScraperRuby::Errors::ScrapeError => error
-    raise error unless error.kind == :invalid_keyword
+    raise(error) unless error.kind == :invalid_keyword
   end
 end
--- a/app/jobs/scrape_keyword_job.rb
+++ b/app/jobs/scrape_keyword_job.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class ScrapeKeywordJob < ApplicationJob
+  queue_as :default
+
+  sidekiq_options retry: 0
+
+  retry_on GoogleScraperRuby::Errors::ScrapeError, wait: 1.minute, attempts: 3 do |job, _error|
+    keyword_id = job.arguments.first[:keyword_id]
+    # TODO(Goose97): better handle database failure
+    # If the below update crashes, the system ends up in an inconsistent state: the job still
+    # has :processing status but already moved to Sidekiq dead queue
+    Keyword.update(keyword_id, status: :failed)
+  end
+
+  def perform(keyword_id:)
+    Google::ScrapeService.new(keyword_id: keyword_id).call!
+  rescue GoogleScraperRuby::Errors::ScrapeError => error
+    raise error unless error.kind == :invalid_keyword
+  end
+end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -11,4 +11,10 @@ class Keyword < ApplicationRecord
             presence: true,
             numericality: { greater_than_or_equal_to: 0 },
             if: :succeeded?
+
+  after_create_commit :enqueue_scrape_job
+
+  def enqueue_scrape_job
+    ScrapeKeywordJob.perform_later keyword_id: id
+  end
 end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -15,8 +15,8 @@ class Keyword < ApplicationRecord
   after_create_commit :enqueue_scrape_job
 
   private
-  
+
   def enqueue_scrape_job
-    ScrapeKeywordJob.perform_later keyword_id: id
+    ScrapeKeywordJob.perform_later(keyword_id: id)
   end
 end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -14,6 +14,8 @@ class Keyword < ApplicationRecord
 
   after_create_commit :enqueue_scrape_job
 
+  private
+  
   def enqueue_scrape_job
     ScrapeKeywordJob.perform_later keyword_id: id
   end
--- a/spec/jobs/scrape_keyword_job_spec.rb
+++ b/spec/jobs/scrape_keyword_job_spec.rb
@@ -13,7 +13,7 @@ describe ScrapeKeywordJob do
             described_class.perform_later(keyword_id: 42)
           end
 
-          assert_performed_jobs 1
+          assert_performed_jobs(1)
         end
       end
 
@@ -29,7 +29,7 @@ describe ScrapeKeywordJob do
 
           perform_enqueued_jobs { described_class.perform_later(keyword_id: keyword.id) }
 
-          assert_performed_jobs 3
+          assert_performed_jobs(3)
         end
 
         context 'when the maximum retry times is exceeded' do
--- a/spec/jobs/scrape_keyword_job_spec.rb
+++ b/spec/jobs/scrape_keyword_job_spec.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe ScrapeKeywordJob do
+  include ActiveJob::TestHelper
+
+  describe '#perform' do
+    context 'given an ERROR occurs' do
+      context "when the keyword doesn't exist" do
+        it 'DOES NOT retry' do
+          perform_enqueued_jobs do
+            described_class.perform_later(keyword_id: 42)
+          end
+
+          assert_performed_jobs 1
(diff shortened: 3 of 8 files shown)


Goose97:
There is still one test left to be implemented: ScrapeKeyword jobs get enqueued after CSV file is uploaded. It's waiting for [this PR](https://github.com/Goose97/google-scraper-ruby/pull/31) to be merged

spec/jobs/scrape_keyword_job_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      context 'when an unexpected error occurs' do
+        it 'retries at most 3 times' do
+          keyword = Fabricate :keyword
+
+          stub = Google::ScrapeService.new(keyword_id: keyword.id)
+          allow(stub).to receive(:call!).and_raise(
+            GoogleScraperRuby::Errors::ScrapeError.new(keyword_id: keyword.id, kind: :unexpected_error)
+          )
+          allow(Google::ScrapeService).to receive(:new).and_return(stub)
```
Goose97: This test makes sure the background job worker handles errors correctly. The above codes is a workaround for `allow_any_instance_of` since that one is not recommended by the community.

However, this is just a hack and basically still do the same thing. 

What is your opinion on this?

app/models/keyword.rb
```diff
@@ -11,4 +11,10 @@ class Keyword < ApplicationRecord
             presence: true,
             numericality: { greater_than_or_equal_to: 0 },
             if: :succeeded?
+
+  after_create_commit :enqueue_scrape_job
+
+  def enqueue_scrape_job
```
andyduong1920, line 19:
```suggestion
  private
  
  def enqueue_scrape_job
```

app/jobs/scrape_keyword_job.rb
```diff
@@ -0,0 +1,21 @@
...
+  queue_as :default
+
+  sidekiq_options retry: 0
+
+  retry_on GoogleScraperRuby::Errors::ScrapeError, wait: 1.minute, attempts: 3 do |job, _error|
+    keyword_id = job.arguments.first[:keyword_id]
+    # TODO(Goose97): better handle database failure
+    # If the below update crashes, the system ends up in an inconsistent state: the job still
+    # has :processing status but already moved to Sidekiq dead queue
+    Keyword.update(keyword_id, status: :failed)
```
andyduong1920, line 13: how about we set it as failed right in the `perform` action 🤔 

```
def perform(keyword_id:)
    Keyword.update(keyword_id, status: :processing)
    Google::ScrapeService.new(keyword_id: keyword_id).call!
  rescue GoogleScraperRuby::Errors::ScrapeError => error
    Keyword.update(keyword_id, status: :failed)
    raise error unless error.kind == :invalid_keyword
  end
```

spec/jobs/scrape_keyword_job_spec.rb
andyduong1920: This is normal @Goose97, I don't see any issue with that :)

app/models/keyword.rb
```diff
@@ -11,4 +11,10 @@ class Keyword < ApplicationRecord
             presence: true,
             numericality: { greater_than_or_equal_to: 0 },
             if: :succeeded?
+
+  after_create_commit :enqueue_scrape_job
```
andyduong1920, line 15: did you check this on your local, technically this branch need to be checkout from the `https://github.com/Goose97/google-scraper-ruby/pull/31` so you could test the whole flow there..

andyduong1920, line 15: We are also missing the unit test for this part - after creating a keyword, it enqueues the job

spec/jobs/scrape_keyword_job_spec.rb
Goose97: cool. I just realize that, while the overall behavior is the same, the `allow_any_instance_of ` is more intrusive. It also stubs existing instances, whereas the above code only stubs newly created instances. TIL

app/models/keyword.rb
Goose97: Yep, I left a note on this PR. Once the #31 is merged, this test will be implemented

Goose97: yeah, I tested against #31 in my local setup. This PR depends on #31, hence the draft status. Once #31 is merged, I will have the whole flow tested again

app/jobs/scrape_keyword_job.rb
Goose97: Oh, I see your point. So the job will be seen as `:failed` in between retry attempts, and marked as `:processing` before each attempt.

My current goal is: only marks a job as `:failed ` if it exceeds the maximum retries count (which is 3). So during the retry period, the status is still `:processing`. This is not an AC so I guess we can relax that.

But even if we go with your suggestion, the issue is still there, am I right? If the update line crashes, there no retry mechanism left to catch the job and it go straight to Sidekiq's dead queue. Anw, this is not a serious issue, I think we can left it like this

olivierobert: It is indeed not a major concern as the UI could have a retry mechanism (out of scope though for the IC). Always aim for covering the happy paths first. Further improvements for edge cases can be handled in subsequent stories.

app/models/keyword.rb
andyduong1920: Next time instead of waiting for the #31 is merged, you can checkout from there directly to have the full flow, what do you think about that idea?

Goose97: wouldn't it make make code reviewing much harder, since we can not isolate changes from this PR 🤔 ?

andyduong1920: No @Goose97 as long as the base branch point to the correct branch which is #31, this PR will display only its changes

image

Goose97: Added test in 32de9e2b3f09716e7fb500a9a6bf522263a955b7

spec/jobs/scrape_keyword_job_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  describe '#perform' do
+    context 'given an ERROR occurs' do
+      context "when the keyword doesn't exist" do
+        it 'DOES NOT retry' do
+          perform_enqueued_jobs do
+            described_class.perform_later(keyword_id: 42)
+          end
+
+          assert_performed_jobs 1
```
olivierobert: MAybe missing parentheses for code consistency?

```suggestion
          assert_performed_jobs(1)
```

spec/jobs/scrape_keyword_job_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        end
+      end
+
+      context 'when an unexpected error occurs' do
+        it 'retries at most 3 times' do
+          keyword = Fabricate(:keyword)
+
+          stub = Google::ScrapeService.new(keyword_id: keyword.id)
```
olivierobert: This syntax can work, but the team usually uses the following to set up stubs/mocks:

```suggestion
          scrape_service = instance_double (Google::ScrapeService)
```

andyduong1920 (approved):

spec/jobs/scrape_keyword_job_spec.rb
Goose97: Fixed in 810d359d3574f9595bdc925c5eb031821d37138a

Not sure why Rubocop lets this passes

spec/jobs/scrape_keyword_job_spec.rb
Goose97: Fixed in 57151f445a6060b23ed73c52de53e51af407d0b9

olivierobert (approved):

Issue #8, opened 2023-07-24
[UI] As a User, I can upload CSV keyword files


Goose97:
Frontend for #9 

## Why

Authenticated users should be able to provide keywords to scrape. To achieve that, we provide an interface for users to upload CSV files.

## Acceptance Criteria

- implement a button for users to upload files
- Should implement filetypes constraint (only allow CSV files)

PR #33, opened 2023-08-09
[UI] As a User, I can upload CSV keyword files

Goose97:
[[UI] As a User, I can upload CSV keyword files](https://github.com/Goose97/google-scraper-ruby/issues/8)

## What happened 👀

Add a button in keyword screen to upload CSV files

## Insight 📝

The current implementation uses Javascript to trigger form submission automatically after a file is selected

## Proof Of Work 📹

https://github.com/Goose97/google-scraper-ruby/assets/35915460/32d4da23-81cf-4787-88ea-86978fce59b5


app/controllers/keywords_controller.rb
```diff
@@ -7,9 +7,9 @@ def create
     form = CsvUploadForm.new
 
     if form.save(create_params[:file])
-      flash[:notice] = I18n.t('csv.upload_success')
+      flash[:success] = I18n.t('activemodel.csv.upload_success')
```
andyduong1920: not sure why we would need to change this @Goose97 as `notice` and `alert` seems like the default Rails way https://api.rubyonrails.org/classes/ActionDispatch/Flash.html

spec/support/file_upload_helpers.rb
```diff
@@ -7,4 +7,12 @@ def self.upload_file(fixture:, content_type: 'text/csv')
       Rack::Test::UploadedFile.new(path, content_type)
     end
   end
+
+  module System
+    def upload_file(fixture)
+      visit(root_path)
+      path = Rails.root.join('spec', 'fixtures', 'files', fixture)
+      page.attach_file('csv_upload_form[file]', path, visible: false)
```
andyduong1920: Just a small thing, as you decided this is the helper which mean can be use for another upload_file field, what do you think if we move the `csv_upload_form[file]` to be a `file_input_field` param, 

```suggestion
      page.attach_file(file_input_field, path, visible: false)
```

spec/systems/csv_upload_form_spec.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'CSV upload form', type: :system do
```
andyduong1920: @Goose97 for the system test, let move this file into the `keywords` folder, so we know this test is about the keyword context

```
spec/systems/keywords/upload_file.rb
```

spec/systems/csv_upload_form_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  describe 'keywords#index' do
+    context 'when user reaches keywords page' do
+      it 'displays the file upload form' do
```
andyduong1920: I think we do not need to test this on the system test layer, as this test is quite simple as we could achieve this in the `requests test`.

- Request test -> Unit test -> fast speed
- UI test -> slow in term of speed

So most of the time, we will use UI test to test only the happy path for the `whole feature`

For this test, I can recommend to just test the happy path

```
it 'redirects to keywords#index path and displayed the success alert' do
end
```

The unhappy path (error) can be nice to have with just 1 error case, no need to test for all error cases on the UI test layer

app/views/keywords/index.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class='screen-keyword'>
```
andyduong1920: this is not the whole keyword screen, this is the main container instead, to detect the whole screen, we normally set the class in the `body` tag in app/views/layouts/application.html.erb

```
<body class="<%= "#{controller_name.dasherize} #{action_name.dasherize}" %>">
```

with that, the body class will have the value per each controller and action for example `body class="keywords index"`

app/views/keywords/_upload_form.erb
```diff
@@ -0,0 +1,14 @@
...
+             method: :post,
+             class: 'form-csv-upload py-4 d-flex justify-content-center'
+            ) do |f| %>
+  <%= f.label(:file, class: 'py-2 px-3') do %>
+    <span class='btn btn-success' role='button'>
+      <%= I18n.t('keywords.upload') %>
+    </span>
+  <% end %>
+
+  <%= f.file_field(:file, class: 'form-csv-upload__file-input d-none', accept: 'text/csv') %>
```
andyduong1920, line 13: I understand you would like to provide a nice UX by hiding the file input here and using some JS to trigger it, hehe, but in terms of the IC; we can go with the more straightforward solution like displaying both `input field and the submit button` there :D no need any JS interaction :D 

For now, the current implement is ok, no need to remove it, but keep that in mind when you implement the other feature, choose the most simpler solution ;)

app/views/keywords/_upload_form.erb
```diff
@@ -0,0 +1,14 @@
+<%= form_for(:csv_upload_form,
+             as: CsvUploadForm,
+             url: keywords_path,
+             method: :post,
+             class: 'form-csv-upload py-4 d-flex justify-content-center'
```
andyduong1920, line 5: Just FYI, in IC we can use the inline CSS class like this, but in our client project, it depends on each project codebase, some projects will not allow to have `inline CSS`, all styles will be written in the CSS components

app/views/keywords/_upload_form.erb
olivierobert: Since it has been implemented, it can stay though 😅 Moving forward let's keep things simple surely.

app/views/keywords/index.html.erb
olivierobert: In case you want to see an example 👉 https://github.com/nimblehq/mia-web/blob/development/app/views/layouts/application.html.slim#L15

olivierobert (commented):
I like your simple UI 💯 Build just what you need in each story 👍

app/views/keywords/_upload_form.erb
Goose97: thanks for the heads up 👍

spec/systems/csv_upload_form_spec.rb
Goose97: Fixed 0a8af24a98a0d9ebd272191f934eaa7de41026f8

app/views/keywords/index.html.erb
Goose97: Fixed in acd5a416897d54627d7bd4d0c3f64c5edb1125a8

app/views/keywords/_upload_form.erb
Goose97: Duly noted. Thanks

spec/systems/csv_upload_form_spec.rb
Goose97: Fixed in 0a8af24a98a0d9ebd272191f934eaa7de41026f8

spec/systems/csv_upload_form_spec.rb
Goose97: Fixed in 39d451a9be45f94cc48dec2601078b5afa674592

spec/support/file_upload_helpers.rb
Goose97: My preferences when it comes to abstraction is: it's better to late than to be early. So I tend to sit on these abstraction opportunity for quite a while 😅 .

But hey, this case seems like a reasonable one. Fixed in 2eb44647871ac2d4d6e3a00a514728c2bb9d84c5

app/controllers/keywords_controller.rb
Goose97: My rationale is to keep the naming consistent with the Bootstrap alert class name. I'm not aware of the conventions around `:notice` and `:alert` till now. Just thought they're some arbitrary field names

Goose97: Fixed in 2490682fdef7a197255898dfcd9832ed00456159

app/views/layouts/application.html.erb
```diff
@@ -8,9 +8,10 @@
 
     <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
     <%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>
-</head>
+  </head>
```
andyduong1920, line 11:
```suggestion
</head>
```

olivierobert (approved):

andyduong1920 (approved):
Just a small comment, the app/javascript/screens/keywords/index.js is following our convention on Compass so all good on that for now 👌

PR #34, opened 2023-08-10
[Backend] As a User, I can view the list of my uploaded keywords

### LINKED ISSUE #11: [Backend] As a User, I can view the list of my uploaded keywords
Backend for #12 

## Why

Users should be able to view their uploaded keywords list

## Acceptance Criteria

- Users should be able to view the list of uploaded keywords

## Notes
- The response will not (yet) contains information about scarping results. These information will be added later once we implement the keyword processing logic

Goose97:
Closes [[Backend] As a User, I can view the list of my uploaded keywords](https://github.com/Goose97/google-scraper-ruby/issues/11)

## What happened 👀

Implement backend for keywords list view

## Insight 📝

- The current implementation uses [pagy](https://github.com/ddnexus/pagy) to handle pagination
- The KeywordQuery class is very bare-bones and contains little to no logics. It will be expand later when we implement advanced features, like search and filter


app/queries/keywords_query.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class KeywordsQuery
+  def initialize(filters = {})
+    # NOTE(Goose97): we don't support any filters for now
+    @filters = filters
```
olivierobert: If it is not needed, then let's remove it 😅 

```suggestion
def initialize()
  #TODO: add filters object as parameter
end
```

Prefer `TODO` over `# NOTE(Goose97):` as it is a standard format usually handled well in text editors (and IDEs).

spec/queries/keywords_query_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+        keywords = described_class.new.call
+
+        expect(keywords).to(be_empty)
+      end
+    end
+
+    context 'given 5 keywords' do
+      it 'returns 5 keywords' do
+        Keyword.delete_all
```
olivierobert: This steps might be overkill as `rspec` reset the database before every test or place db operations in transactions that are rolled back between tests 💡

app/queries/keywords_query.rb
Goose97: Fixed in d93acb0abbf3765eceda45d0cc3316b982af5365

spec/queries/keywords_query_spec.rb
Goose97: I was wondering if that violates this guideline https://nimblehq.co/compass/development/code-conventions/ruby/rspec/#incidental-state 🤔

spec/queries/keywords_query_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  describe '#call' do
+    context 'given NO keywords' do
+      it 'returns an empty list' do
+        keywords = described_class.new.call
+
+        expect(keywords).to(be_empty)
+      end
+    end
+
+    context 'given 5 keywords' do
```
olivierobert: Such test can be misleading as it should work for any number of keywords:

```suggestion
    context 'given a list of keywords' do
```

spec/queries/keywords_query_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      it 'returns 5 keywords' do
+        Fabricate.times(5, :keyword)
```
olivierobert: Always create the minimum of fixtures. A list starts at 2 so create only two keywords. More would be needed if we wanted to test the pagingation.

spec/queries/keywords_query_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    context 'given NO keywords' do
+      it 'returns an empty list' do
+        keywords = described_class.new.call
+
+        expect(keywords).to(be_empty)
+      end
+    end
+
+    context 'given 5 keywords' do
+      it 'returns 5 keywords' do
```
olivierobert:
```suggestion
      it 'returns all keywords' do
```

spec/support/webdrivers.rb
```diff
@@ -22,3 +22,5 @@
   ]
   WebMock.disable_net_connect!(allow_localhost: true, allow: allowed_sites)
 end
+
+Webdrivers::Chromedriver.required_version = '114.0.5735.90'
```
olivierobert: I do not recommend setting the version explicitly unless this sets a minimum version. Chrome has a six-week release cycle, so we would need to change that bit often.

 In practice, devs need to regularly update their local Chromedriver binary as newer Chrome releases do not support older versions.

app/controllers/keywords_controller.rb
```diff
@@ -1,13 +1,20 @@
 # frozen_string_literal: true
 
 class KeywordsController < ApplicationController
-  def index; end
+  def index
+    pagy, keywords = pagy(KeywordsQuery.new.call)
+
+    render(locals: {
+             pagy: pagy,
+             presenters: keywords
```
olivierobert: Avoid plurals and be more descriptive (as there could be more than one presenter):

```suggestion
             keyword_presenter: keyword_presenter
```

Goose97: The presenter is implemented in the frontend PR. So I'll just remove this line for now.

Fixed in d36d4221b027746402452deaea288f98810af599

spec/queries/keywords_query_spec.rb
Goose97: Fixed in e5f09346cd54e1b94fc42bbaa94df8637871bd91

spec/queries/keywords_query_spec.rb
Goose97: Fixed in e5f09346cd54e1b94fc42bbaa94df8637871bd91

spec/queries/keywords_query_spec.rb
Goose97: Fixed in 4f4526bea3a181ad4a5dc98745b8aeb95387ab9f

spec/support/webdrivers.rb
Goose97: Fixed in 7a85b3a20bdf4ace392c44f8fc65b3e521d4e3ce

Installed chromedriver version 115.0.5790.170 from https://googlechromelabs.github.io/chrome-for-testing/#stable

spec/queries/keywords_query_spec.rb
olivierobert: No, because Rails handles it for us :-)

olivierobert (approved):

olivierobert:
To unblock you and move forward, @Goose97 let's merge this PR even with only one approval since Andy is on leave.

PR #36, opened 2023-08-14
[UI] As a User, my uploaded are processed immediately

### LINKED ISSUE #16: [UI] As a User, my uploaded keywords are processed immediately
## Why

Processing upload keywords immediately provides a smooth and snappier interface, enhances the application UX

## Acceptance Criteria

- After upload a file successfully, the UI is updated to reflect the new scraping job
- While the file is processed, the UI is updated in real-time to reflect the current status. For each keyword, it should display:
  - The current status: pending / processing / succeeded / failed
  - A link to see details information once the scraping succeeds

## Notes
- ~~`retry keywords` are keywords which have failed to scrape and have been scheduled for retry~~ (We don't support retry status for now). `error keywords` are keywords which have exceeded the maximum retry times. See #15 for more details

Goose97:
Closes [[UI] As a User, my uploaded keywords should be processed immediately](https://github.com/Goose97/google-scraper-ruby/issues/16)

## What happened 👀

Implement real-time UI updates for list keywords view

## Insight 📝

- Using turbo-rails to perform push updates.
- When keywords finish processing (succeeded or failed), the UI is updated accordingly.

## Proof Of Work 📹

I added a 5 seconds delay in the demo video to clearly demonstrate the status transitions
https://github.com/Goose97/google-scraper-ruby/assets/35915460/e84be3b5-7c3a-473f-9ecf-b4f9617c4324


app/models/keyword.rb
```diff
@@ -11,4 +11,25 @@ class Keyword < ApplicationRecord
...
+  after_create_commit :enqueue_scrape_job
+  after_update_commit :broadcast_status_update
+
+  private
+
+  def enqueue_scrape_job
+    ScrapeKeywordJob.perform_later(keyword_id: id)
+  end
+
+  def broadcast_status_update
```
olivierobert: While Rails makes it powerful to use model hooks to trigger actions, it is a tool we should use sparingly as 1) it is hard to test and 2) it mixes too much business logic (this use cases) with business entities. Moving this kind of use case to the form or service classes is better than placing them in the models. It si nothing wrong per se but it is a better application architecture choice.

To recap, using `after_create_commit` is fair use, but it expected business logic for all keywords.  Using `after_update_commit` to broadcast an update to the UI would be better in the service class `Google::ScrapeService`, especially as it is odd for a model to call a presenter class 🤔 (only decorators or job classes are usually called in models).

app/views/keywords/_keywords_table.erb
```diff
@@ -12,17 +14,7 @@
 
     <tbody class="keywords-table__body">
       <% keyword_presenter.map do |presenter| %>
```
olivierobert: Rails has built-in ways to render collections avoiding to use of `map`. See: `https://thoughtbot.com/blog/rendering-collections-in-rails`. In addition, to render collection, it is best to have a collection and item presenter. See https://github.com/nimblehq/mia-web/blob/development/engines/mia_campaign/app/presenters/mia_campaign/campaign_collection_presenter.r,  https://github.com/nimblehq/mia-web/blob/development/engines/mia_campaign/app/presenters/mia_campaign/campaign_presenter.rb and https://github.com/nimblehq/mia-web/blob/752f45a860b25bdc456c1bcc82b6b5a6ca18e334/engines/mia_admin/app/controllers/mia_admin/campaigns_controller.rb#L17.

So we could do the following:

```ruby
<%= keyword_list_presenter.keywords %>
```

olivierobert (changes requested):

app/models/keyword.rb
Goose97: Fixed in d1849b45b94658f7ff80415f783db57423b27cb4

Good point. I refactored the status broadcasting logic in to the ScraperService. 

But I'm wondering about the situation where another service class needs to broadcast keyword updates. Should it duplicate this logic in their own class. What do you think about extracting it to a `KeywordBroadcastable` concern? Obviously this is overkill for the current situation.

app/views/keywords/_keywords_table.erb
Goose97: Fixed in 8cff1531c0df3753bdd71b255eaf1fbab8f09ee6

app/models/keyword.rb
olivierobert: Concerns as like mixins/modules so yes that would be the way to go not to duplicated code 👍

spec/fabricators/keyword_fabricator.rb
```diff
@@ -4,3 +4,11 @@
   content { FFaker::Lorem.characters(16) }
   status :processing
 end
+
+Fabricator(:parsed_keyword, from: :keyword) do
```
olivierobert, line 8: Nice use case for a separate fixture 👍

olivierobert (approved):
Looks great :-)

app/presenters/keyword_presenter.rb
```diff
@@ -14,11 +14,19 @@ def initialize(keyword:)
     @keyword = keyword
   end
 
+  def status
+    keyword.status.capitalize
```
andyduong1920: Just a small thing, should we localize the status :)

spec/systems/keywords/list_keywords_spec.rb
```diff
@@ -13,19 +15,34 @@
...
+    it 'updates the UI after the keyword finish processing', vcr: 'google/google_no_ads' do
+      keyword = Fabricate(:keyword)
+
+      visit(root_path)
+      perform_enqueued_jobs do
+        ScrapeKeywordJob.perform_later(keyword_id: keyword.id)
+      end
+
+      within("tbody > tr[data-keyword-id=\"#{keyword.id}\"]") do |tr|
+        expect(tr.find('td[data-testid="status"]')).to(have_content('Succeeded'))
```
andyduong1920: Let's localize the `Succeeded`

andyduong1920 (approved):
Just 2 small comments about the status localization @Goose97

andyduong1920:
Also fix the Lint issue too @Goose97 

image

spec/systems/keywords/list_keywords_spec.rb
Goose97: Fixed in 9cdf12a0d27bd886e7cc3bd5248a02835b4523b0

app/presenters/keyword_presenter.rb
Goose97: Fixed in 9cdf12a0d27bd886e7cc3bd5248a02835b4523b0

PR #37, opened 2023-08-16
[API] As a User, I can view the list of my uploaded keywords

### LINKED ISSUE #18: [API] As a User, I can view the list of my uploaded keywords
## Why

Allowing third-party applications to interact with us

## Acceptance Criteria

- Provide an API to fetch users uploaded keywords
- Pagination must be implemented
- Do NOT implement authentication. This will be added later once we set up the authentication pipeline

Goose97:
Closes [[API] As a User, I can view the list of my uploaded keywords](https://github.com/Goose97/google-scraper-ruby/issues/18)

## What happened 👀

Implement list keywords API

## Insight 📝

- The response item does not include all attributes from keyword object. Some attributes only available through view keyword's details API
- The current implementation uses [jsonapi-serializer](https://jsonapi-rb.org/) to perform serialization
- The endpoint for the API is `GET /api/v1/keywords`


spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe(Api::V1::KeywordsController) do
```
Goose97, line 5: Should I add some tests to assert we return correct response data?

andyduong1920:
@Goose97 let's follow our PR convention :) https://nimblehq.co/compass/development/code-reviews/author-guidelines/

image

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class KeywordsController < ApplicationController
+      include ErrorHandlerConcern
+
+      def index
+        _, keywords = pagy(KeywordsQuery.new.call)
+        render(json: KeywordSerializer.new(keywords))
```
andyduong1920:
```suggestion

        render(json: KeywordSerializer.new(keywords))
```

app/serializers/keyword_serializer.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class KeywordSerializer
```
andyduong1920: How about adding the `V1` here as well @Goose97

spec/requests/api/v1/keywords_controller_spec.rb
andyduong1920: yes @Goose97

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class KeywordsController < ApplicationController
```
olivierobert, line 5: How about defining a new base controller for the API:

```suggestion
    class KeywordsController < API::ApplicationController
```

Then `rescue Pagy::OverflowError => error` could be defined at that base controller for all endpoints 💡 

See https://github.com/nimblehq/the1-api/blob/development/app/controllers/v2/application_controller.rb#L16 fr examples on how to handle application-wide errors

spec/requests/api/v1/keywords_controller_spec.rb
olivierobert: The team usually uses https://github.com/thoughtbot/json_matchers. See this example: https://github.com/nimblehq/the1-api/blob/development/spec/controllers/the1_badge/v2/badges_controller_spec.rb#L16

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        page_size = Pagy::DEFAULT[:items]
+        Fabricate.times(page_size, :keyword)
+
+        get(api_v1_keywords_path, params: { page: 1 })
+
+        expect(response).to(have_http_status(:success))
+      end
+    end
+
+    context 'given a INVALID page params' do
```
olivierobert: English issue:

```suggestion
    context 'given INVALID page params' do
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        page_size = Pagy::DEFAULT[:items]
+        Fabricate.times(page_size + 1, :keyword)
+
+        get(api_v1_keywords_path)
+
+        expect(json_response[:data].count).to(eq(page_size))
+      end
+    end
+
+    context 'given a VALID page params' do
```
olivierobert, line 16: English issue:

```suggestion
    context 'given VALID page params' do
```

app/serializers/keyword_serializer.rb
Goose97: Fixed in 4b9c59c46f32fe39ec662e53caa1abf4d99a2d71

spec/requests/api/v1/keywords_controller_spec.rb
Goose97: Fixed in 2b5e7442961bd7d26566fe5861be0fd2028457bf

app/controllers/api/v1/keywords_controller.rb
Goose97: This is indeed better 👍 . Fixed in 38a90735bf1cac906131759d368498838ff8cfbe

spec/requests/api/v1/keywords_controller_spec.rb
Goose97: Fixed in 498ef53ece2e7c66aad601d42f9534e2fa532b36

app/controllers/concerns/api/v1/error_handler_concern.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module ErrorHandlerConcern
```
olivierobert: Naming-wise, adding the suffix `concern` is not needed. Concerns are usually prefixed with `has` or suffixed with `able` (see https://dev.to/mainstreet/how-to-name-rails-concerns-3m86). In this case, we might not even need a concern since the code is short and not shared, so it could be placed in `ApplicationController < ActionController::API`. This way all error handling, including the pagy rescue block, would be in the same file.

app/serializers/v1/keyword_serializer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module V1
+  class KeywordSerializer
+    include JSONAPI::Serializer
```
olivierobert: It would be best to create a base serailizer and make that class inherit from it not to have to include that module in each serializer 💡

olivierobert (commented):
Just some minor changes needed and this would be ready to go 🚀

app/serializers/v1/keyword_serializer.rb
Goose97: Fixed in b55c8f986c755e4cfa50f22f26e118f1474428fb

app/controllers/concerns/api/v1/error_handler_concern.rb
Goose97: Fixed in 8ed84408013de8ec98bb335c5f3f7661cf297acc

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+      def index
+        _, keywords = pagy(KeywordsQuery.new.call)
+
+        render(json: ::V1::KeywordSerializer.new(keywords))
```
andyduong1920, line 9: Just a minor thing, I'm curious what error if we do not add the `::` in front of the `V1::KeywordSerializer` 

```suggestion
        render(json: V1::KeywordSerializer.new(keywords))
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe(Api::V1::KeywordsController) do
+  describe 'GET #index' do
+    context 'given one page of keywords' do
+      it 'returns all keywords' do
```
andyduong1920: i think we do not need to test just, just test the `it 'returns the first page of keywords' do` is enough

andyduong1920 (approved):
Just 1 small comment and 1 question @Goose97 👍

spec/requests/api/v1/keywords_controller_spec.rb
Goose97: Does this mean reduces these test cases, from:
```
context 'given one page of keywords' do
  it 'returns all keywords' do
  end
end

context 'given more than one page of keywords' do
  it 'returns the first page of keywords' do
  end
end
```

to only:
```
it 'returns the first page of keywords' do
end
```

app/controllers/api/v1/keywords_controller.rb
Goose97: It'll raise this error (I still don't know why)

```
NameError:
  uninitialized constant Api::V1::KeywordSerializer
```

Should I wrap the serializer under the Api namespace as well, change it from:
```ruby
module V1
  class KeywordSerializer
  end
end
```

to

```ruby
module Api
  module V1
    class KeywordSerializer
    end
  end
end
```

andyduong1920: Ah I see, then let keep using the `::V1::KeywordSerializer.new(keywords)` @Goose97 👍

spec/requests/api/v1/keywords_controller_spec.rb
andyduong1920: Yeah @Goose97, what do you think?

olivierobert (approved):

spec/requests/api/v1/keywords_controller_spec.rb
Goose97: I think it's fine to remove those tests. Fixed in 158a1be6c9b3341f7f14c9e57387f4df4ffef04c

PR #38, opened 2023-08-21
[Backend] As a User, I can view the details of my keywords

### LINKED ISSUE #21: [Backend] As a User, I can view the details of my keywords
## Why

The keyword list view only show basic details about the keyword. Users need a way to see the full details of a given keyword

## Acceptance Criteria

- Implement a separate view to see keyword details
- The view MUST provide:
  - Number of AdWords advertisers in the top position
  - Total number of AdWords advertisers on the page
  - URLs of the AdWords advertisers in the top position
  - Number of the non-AdWords results on the page
  - URLs of the non-AdWords results on the page
  - Total number of links (all of them) on the page
  - A view of the page (using the cached HTML in #14)

Goose97:
Closes [[Backend] As a User, I can view the details of my keywords](https://github.com/Goose97/google-scraper-ruby/issues/21)

## What happened 👀

Prepare backend for show keyword details view

## Insight 📝

- Implement KeywordDecorator responsible for querying database and displaying information about a keyword
- Implement query objects responsible for querying the number of top ads entries, the number of ads entries, and the number of non ads entries

## Proof Of Work 📹

Show us the implementation: screenshots, GIFs, etc.


app/models/keyword.rb
```diff
@@ -4,6 +4,10 @@ class Keyword < ApplicationRecord
   enum status: { pending: 'pending', processing: 'processing', succeeded: 'succeeded', failed: 'failed' }
 
   has_many :keyword_search_entries, dependent: :destroy
+  has_one :keyword_stats, dependent: :nullify
+
+  delegate :top_ads_urls, :non_ads_urls, to: :keyword_decorator
+  delegate :top_ads_count, :total_ads_count, :non_ads_count, to: :keyword_stats, allow_nil: true
```
Goose97: The current behaviour is: if the `keyword` doesn't have any search entries, `keyword_stats` would be nil. All delegates fields as a result would be nil as well

```ruby
# This keyword doesn't have any search entries
keyword = Keyword.find(1)

keyword.keyword_stats # nil

keyword.keyword_stats.top_ads_count # nil
```

The behavior I want in the above case is: `keyword.keyword_stats.top_ads_count` equals 0 instead of nil. How can I achieve this?

What I've tried so far:
1. Manually setup `top_ads_count` method:
```ruby
def top_ads_count
  return keyword_stats ? keyword_stats.top_ads_count : 0
end
```

2. In case `keyword_stats` is nil, fallback to an empty instance
```ruby
delegate ..., to: keyword_stats_with_fallback

def keyword_stats_with_fallback
  return keyword_stats || KeywordStats.new() # we need some mechanisms to populate fields here as 0, otherwise they're all nil
end
```

What is the preferred way or is there any other solutions?

olivierobert: Technically, the API response only needs the value set to zero. At the model layer, having `nil` is not an issue per se. So how about handling the data formatting at the serializer level instead? (which act as a view logic layer)

db/views/keyword_stats_v01.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+  keyword_id,
+  COUNT(*) FILTER (WHERE kind = 'ads' AND position = 'top') as top_ads_count,
+  COUNT(*) FILTER (WHERE kind = 'ads') as total_ads_count,
+  COUNT(*) FILTER (WHERE kind = 'non_ads') as non_ads_count
```
olivierobert: While scenic is the right choice to implement DB views, the current use case is relatively too simple to bring in such a gun 😅 Showing the total can also be achieved by a query object (see https://github.com/nimblehq/mia-web/tree/752f45a860b25bdc456c1bcc82b6b5a6ca18e334/engines/mia_admin/app/queries/mia_admin). 

Prefer to use DB views for complex reporting features 💡

olivierobert (commented):
I recommend simplifying the implementation as it is relatively complex to display simple enough to compute data. Factor in that the associated data is not that huge so there is no need to think about performance much.

Gemfile
```diff
@@ -16,6 +16,7 @@ gem 'nokogiri' # Work with XML and HTML from Ruby
 gem 'turbo-rails' # Provide SPA-like feature without writing Javascript
 # gem 'jbuilder' # Build JSON APIs with ease
 gem 'redis' # Use Redis adapter to run Action Cable in production
+gem 'scenic' # Create and manage database views in Rails
```
andyduong1920: @Goose97 can we just go with the simplest solution by loading overall `KeywordSearchEntry` entries instead of using the DB views for this, as Olivier mentioned, DB Views seem like an overkill for this  case, we could use Query object or for me, loop over the `KeywordSearchEntry` records, process the data by using Ruby :) 

```
def top_ads_urls
  keyword_search_entries.top.ads.pluck(:url)
end

...
```

app/models/keyword.rb
Goose97: There's definitely no problem handling that at serializer/presenter layers. However, I think it would be best if the can handle that in the model/decorator layer, to simplify the interface.

But in this case, there seems to be no easy way to achieve this. So I'll go with your suggestion.

db/views/keyword_stats_v01.sql
Goose97: Got it. The reason I go with view in the first place is that it reduces the amount of database queries. However, factoring in the added complexity, it doesn't seem to worth the trade off.

I will go with query objects approach

Goose97: Implement query object in 91105728caa1dda513ea936fc843ecd8690575a5

Gemfile
Goose97: Implement query object in 91105728caa1dda513ea936fc843ecd8690575a5

app/models/keyword.rb
```diff
@@ -4,6 +4,9 @@ class Keyword < ApplicationRecord
   enum status: { pending: 'pending', processing: 'processing', succeeded: 'succeeded', failed: 'failed' }
 
   has_many :keyword_search_entries, dependent: :destroy
+  has_one :keyword_stats, dependent: :nullify
```
olivierobert: How will this be used? I am unsure eif this is needed 🤔

app/models/keyword.rb
```diff
@@ -4,6 +4,9 @@ class Keyword < ApplicationRecord
...
+
+  delegate :top_ads_count, :total_ads_count, :non_ads_count, :top_ads_urls, :non_ads_urls, to: :keyword_search_entries_query
```
olivierobert: It is a bit unexpected to delegate to a query object 🤔 It works, and to some extent, it makes sense, but usually, the only objects interacting with models are decorations 💡 

An alternative is not coupling the query in the model and limiting the usage to the controller 💡 

> [!NOTE]
> There is also https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-extending to add new query methods

app/models/keyword.rb
Goose97: Oops, forgot to remove that 🤕 . Fixed in e8e11f81b5d7df46200e5b567573eccb30b54a62

app/models/keyword.rb
Goose97: Your suggestion does sound better. I remove the delegation logic in 5cb853aa0b9475d28e5e71fd1ccbb5927bb13403

The logic will be move to controller in the frontend PR

app/models/keyword_stats.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class KeywordStats < ApplicationRecord
```
andyduong1920: do we still need this file @Goose97 🤔

app/queries/keyword_search_entries_query.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def with_keyword(keyword_id)
+    @relation = relation.where(keyword_id: keyword_id)
+
+    self
+  end
+
+  def top_ads_count
+    relation.where(kind: :ads, position: :top).size
```
andyduong1920: How about using the `COUNT in SQL` directly, if yes, let change for another methods as well 👍 

```suggestion
    relation.count(kind: :ads, position: :top)
```

app/queries/keyword_search_entries_query.rb
```diff
@@ -0,0 +1,41 @@
...
+  def total_ads_count
+    relation.where(kind: :ads).size
+  end
+
+  def non_ads_count
+    relation.where(kind: :non_ads).size
+  end
+
+  def top_ads_urls
+    urls_of(relation.where(kind: :ads, position: :top))
```
andyduong1920: is it duplicated with the KeywordDecorator.rb file there 🤔

app/queries/keyword_search_entries_query.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class KeywordSearchEntriesQuery
+  def initialize(relation = KeywordSearchEntry)
+    @relation = relation
+  end
+
+  def with_keyword(keyword_id)
```
andyduong1920: How about

```suggestion
  def by_keyword_id(keyword_id)
```

spec/queries/keyword_search_entries_query_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+      it 'returns an empty array' do
+        keyword = Fabricate(:parsed_keyword) { keyword_search_entries(count: 0) }
+
+        query = described_class.new.with_keyword(keyword.id)
+
+        expect(query.non_ads_urls).to(be_empty)
+      end
+    end
+
+    context 'given a list of non ads search entries' do
```
andyduong1920, line 119: in general we put the happy case first, so the `context 'given a list of non ads search entries' do` first, `context 'given NO non ads search entries' do` later

spec/queries/keyword_search_entries_query_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+        Fabricate(
+          :keyword_search_entry,
+          kind: :non_ads,
+          keyword_id: keyword.id,
+          urls: ['https://www.ruby-lang.org/en/']
+        )
+
+        query = described_class.new.with_keyword(keyword.id)
+
+        expect(query.non_ads_urls).to(contain_exactly('https://google.com', 'https://bing.com', 'https://www.ruby-lang.org/en/'))
```
andyduong1920, line 140: As the implementation has the uniq `query.pluck(:urls).flatten.uniq` so let's have the test to cover that too @Goose97

app/queries/keyword_search_entries_query.rb
olivierobert: My two cents: the following are all equivalent to me 😅 

```suggestion
  def with_keyword(keyword_id)
```

```suggestion
  def by_keyword(keyword_id)
```

```suggestion
  def for_keyword(keyword_id)
```

The key is to be consistent throughout the codebase 💡

app/queries/keyword_search_entries_query.rb
olivierobert: We could bypass using `count` if the associated records are already loaded. See the difference between `length`, `count`, and `size` 👉 https://stackoverflow.com/a/6083229

app/decorators/keyword_decorator.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class KeywordDecorator < SimpleDelegator
```
olivierobert: Is this decorator needed as it does not seem used?

olivierobert (commented):
I just added some context to the issues raised by Andy. This PR seems to include some left-over code from the previous implementation 🙈

spec/queries/keyword_search_entries_query_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+        )
+
+        Fabricate(
+          :keyword_search_entry,
+          kind: :non_ads,
+          keyword_id: keyword.id,
+          urls: ['https://www.ruby-lang.org/en/']
+        )
+
+        query = described_class.new.with_keyword(keyword.id)
```
andyduong1920: Also we seem to test only with the `with_keyword` filter, should we have the test without that? 

```
query = described_class.new

.....
```

app/queries/keyword_search_entries_query.rb
andyduong1920: Ah, thanks @olivierobert then `size` will be the right solution here 👍

app/queries/keyword_search_entries_query.rb
Goose97: My bad. I forgot to remove that file. Fixed in 6310a736cc744121a554172071b42970e95a8e83

app/models/keyword_stats.rb
Goose97: Removed in ce0cfdcdc4cc60406a727391eeaffee341e80b62

app/decorators/keyword_decorator.rb
Goose97: Removed in 6310a736cc744121a554172071b42970e95a8e83

spec/queries/keyword_search_entries_query_spec.rb
Goose97: I just made keyword_id to be a required parameter here 592ac6c25fa3727924b5a9f6d57aa5a7591ec210. It doesn't make sense for our app to query without a keyword. So I think we can omit this test case.

spec/queries/keyword_search_entries_query_spec.rb
Goose97: Good call. Fixed in 516d31198e7a65ad29e5ea71c0c01e4066482a34

spec/queries/keyword_search_entries_query_spec.rb
Goose97: Fixed in 4e2c448fb7a53aaac846d9b2075bfd19f7883f8c

olivierobert (approved):

app/queries/keyword_search_entries_query.rb
```diff
@@ -0,0 +1,36 @@
...
+  def non_ads_count
+    relation.where(kind: :non_ads).size
+  end
+
+  def top_ads_urls
+    urls_of(relation.where(kind: :ads, position: :top))
+  end
+
+  def non_ads_urls
+    urls_of(relation.where(kind: :non_ads))
```
andyduong1920: here we call the `relation.where(kind: :non_ads)` 2 times from 2 different methods, we know the Active record will cache the query result on their side, but to make the code more clear, and also caching it on the Ruby side by using [||= operator](https://stackoverflow.com/questions/995593/what-does-or-equals-mean-in-ruby), should we refactor it like

image

```
def non_ads_count
   non_ads.size
 end

def non_ads_urls
  urls_of(non_ads)
end

private

def non_ads
  @non_ads ||= relation.where(kind: :non_ads)
end

# Same for top_ads
def top_ads
  @top_ads ||= relation.where(kind: :ads, position: :top)
end
```

andyduong1920 (approved):
Just a small comment @Goose97

app/queries/keyword_search_entries_query.rb
Goose97: Nice suggestion. Added in 45cf7722527dfd3fa1f64acb540c5fb57ef43355

PR #39, opened 2023-08-23
[UI] As a User, I can view the details of my keywords

### LINKED ISSUE #22: [UI] As a User, I can view the details of my keywords
Frontend for #21

## Why

The keyword list view only show basic details about the keyword. Users need a way to see the full details of a given keyword

## Acceptance Criteria

- Users MUST have a way to navigate to the detailed view from the keyword list view
- Users MUST have a way to navigate to the keyword list view from the detailed view

Goose97:
Closes [[UI] As a User, I can view the details of my keywords](https://github.com/Goose97/google-scraper-ruby/issues/22)
## What happened 👀

Implement keyword details view

## Insight 📝

- Display all keyword statistics in #21 
- Display view of the search result page
- The view has a back to home page button

## Proof Of Work 📹

Screenshot 2023-08-23 at 10 50 15

image


app/assets/stylesheets/screens/_keyword_details.scss
```diff
@@ -0,0 +1,6 @@
+.keyword-details {
```
olivierobert: For screens, the body class (see in https://github.com/Goose97/google-scraper-ruby/blob/develop/app/views/layouts/application.html.erb#L13) name must be used:

```suggestion
.keywords.show {
```

We also missed that in https://github.com/Goose97/google-scraper-ruby/blob/develop/app/assets/stylesheets/screens/_keywords.scss#L1 where is should be:

```
.keywords.index {
```

> [!NOTE]
> We could combine all keyword-related screens into one file: `_keywords.scss` then have multiple sections:
> ```
> .keywords.index {}
> 
> .keywords.show {}
> ```

app/controllers/keywords_controller.rb
```diff
@@ -10,6 +10,21 @@ def index
            })
   end
 
+  def show
+    keyword = Keyword.find(params['id'])
+
+    search_entries_query = KeywordSearchEntriesQuery.new.with_keyword(keyword.id)
```
olivierobert: When looking at its usage, it feels more natural to do:

```suggestion
    search_entries_query = KeywordSearchEntriesQuery.new(keyword_id: keyword.id)
```

app/assets/stylesheets/screens/index.scss
```diff
@@ -1 +1,2 @@
-@import './keywords'
+@import './keywords';
+@import './keyword_details';
```
andyduong1920: As we have the controller as the action on the body tag (`<body class="<%= "#{controller_name.dasherize} #{action_name.dasherize}" %> px-4 py-2">`), we no need to separate it into 2 files for the screen @Goose97, instead we can have only 1 `screens/_keywords.scss` and the content for that separate per action

```
body.keywords {
  &.index {
    .....
  }

  &.show {
   ....
  }
```

app/controllers/keywords_controller.rb
```diff
@@ -10,6 +10,21 @@ def index
            })
   end
 
+  def show
+    keyword = Keyword.find(params['id'])
```
andyduong1920: Just a small thing we can the `:` in this case ;) 
```suggestion
    keyword = Keyword.find(params[:id])
```

spec/systems/keywords/view_keyword_detail_spec.rb
olivierobert, line 1: We could skip the suffix `_detail` as it seems redundant if we have the following tests:

```
- view_keyword_list_spec.rb
- view_keyword_spec.rb
```

Based on the name, we can understand one is for the `index` action and the other for the `show` action.

app/controllers/keywords_controller.rb
```diff
@@ -10,6 +10,21 @@ def index
...
+
+    search_entries_query = KeywordSearchEntriesQuery.new.with_keyword(keyword.id)
+
+    render(locals: {
+             keyword: keyword,
```
andyduong1920, line 19: how about just passing the `search_entries_query` @Goose97 
```suggestion
             search_entries_query: search_entries_query
```

config/locales/en.yml
```diff
@@ -8,3 +8,4 @@ en:
       processing: "Processing"
       succeeded: "Succeeded"
       failed: "Failed"
+    to_home_page_link: "Back to home page"
```
olivierobert: This might be used on more thank one page so moving to shared section could be good:

```yaml
en:
  keywords:
    #...
  shared:
    link:
      back_to_home: "Back to home page"
```

app/views/keywords/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  <p data-testid="keyword_details_ads_urls">
+    <%= render('url_list', urls: top_ads_urls, title: "Top ads URLs") %>
+  </p>
+
+  <p data-testid="keyword_details_non_ads_urls">
+    <%= render('url_list', urls: non_ads_urls, title: "Non ads URLs") %>
+  </p>
+
+  <div data-testid="keyword_details_page_view">
+    <p>Page view:</p>
```
andyduong1920: As we use the `I18n.t` in other place, let keep it consistent @Goose97, let use the. I18n.t here as well, just a small thing

app/views/keywords/show.html.erb
```diff
@@ -0,0 +1,24 @@
+<a href="<%= keywords_path %>">
+  <%= I18n.t('keywords.to_home_page_link') %>
+</a>
+
+<div class="d-flex flex-column justify-content-center mt-4">
+  <p data-testid="keyword_details_content">Keyword: <%= keyword.content %></p>
+  <p data-testid="keyword_details_links_count">Total links: <%= keyword.links_count %></p>
+  <p data-testid="keyword_details_top_ads_count">Top ads count: <%= top_ads_count %></p>
+  <p data-testid="keyword_details_total_ads_count">Ads count: <%= total_ads_count %></p>
+  <p data-testid="keyword_details_non_ads_count">Non-ads count: <%= non_ads_count %></p>
```
andyduong1920:
```suggestion
  <p data-testid="keyword_details_non_ads_count">I18n.t('xxx'): <%= non_ads_count %></p>
```

spec/systems/keywords/view_keyword_detail_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'View keyword details', type: :system do
+  include ActiveJob::TestHelper
```
andyduong1920: Do we need this 🤔 

```suggestion
```

spec/systems/keywords/view_keyword_detail_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  # rubocop:disable RSpec/ExampleLength
+  it 'displays a link to keywords#show for succeeded keyword and displays the pagination nav' do
```
olivierobert: The test does not describe what it does, as there are many assertions for the scraped content. The link to the homepage is a minor detail.

```suggestion
  it 'displays the scraped content information' do
```

spec/systems/keywords/view_keyword_detail_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    visit(keyword_path(keyword))
+
+    expect(page).to(have_selector('p[data-testid="keyword_details_content"]'))
+    expect(page).to(have_selector('p[data-testid="keyword_details_links_count"]'))
+    expect(page).to(have_selector('p[data-testid="keyword_details_top_ads_count"]'))
+    expect(page).to(have_selector('p[data-testid="keyword_details_total_ads_count"]'))
+    expect(page).to(have_selector('p[data-testid="keyword_details_non_ads_count"]'))
+    expect(page).to(have_selector('p[data-testid="keyword_details_ads_urls"]'))
+    expect(page).to(have_selector('p[data-testid="keyword_details_non_ads_urls"]'))
+    expect(page).to(have_selector('div[data-testid="keyword_details_page_view"]'))
```
olivierobert: Only checking for having the selectors is not sufficient. The test would still pass if the scraped data was empty. We must ensure that the scraped content is displayed, i.e., checking the text value in each DOM selector.

olivierobert (changes requested):

app/assets/stylesheets/screens/index.scss
Goose97: Good call. Fixed in dfe9a4bb41f9e6af9189f0174273912661b66ffe

app/controllers/keywords_controller.rb
Goose97: Yeah, that's what I've been wondering too. One one hand, it does make the code less verbose. But it also add more responsibility for the view, namely querying the database.

In my mind, views should be pure and dumb. It only receives inputs as plain data and outputs a HTML / JSON payload. That might be over thinking for this case though 😄

app/views/keywords/show.html.erb
Goose97: Agree. Fixed in b99101be0f14e8d84cf2e88f9d68a836a980d681

app/views/keywords/show.html.erb
Goose97: Fixed in b99101be0f14e8d84cf2e88f9d68a836a980d681

spec/systems/keywords/view_keyword_detail_spec.rb
Goose97: Good catch. Forgot to delete it. Fixed in 8e16ef2f310b539f7d5c37fa9ceee7138d219f69

app/assets/stylesheets/screens/_keyword_details.scss
Goose97: Fixed in dfe9a4bb41f9e6af9189f0174273912661b66ffe as Andy suggested

config/locales/en.yml
Goose97: Agree. Fixed in d483681c6a18bb3da74436a75d30e434f37cf319

spec/systems/keywords/view_keyword_detail_spec.rb
Goose97: Test case improved in 9be9a361e317cafab5bb7ab0eb04e3b714f1581f

spec/systems/keywords/view_keyword_detail_spec.rb
Goose97: Fixed in 9be9a361e317cafab5bb7ab0eb04e3b714f1581f

spec/systems/keywords/view_keyword_detail_spec.rb
Goose97: Agree. Fixed in e6dc082038b056fb077daa286df1f55404ee912d

app/controllers/keywords_controller.rb
Goose97: 👍 . Fixed in 468e19e89260ce8b9e9411c38ba36241d1ff1bb8

app/controllers/keywords_controller.rb
olivierobert: The view would not do the querying as this is handle by the query object. The controller invokes the query objects and then pass an object to the view (similar to `assigns` in Phoenix) so I do not see any concerns here :-)

app/controllers/keywords_controller.rb
Goose97: Fixed in afc2ef794bdd0ccd4ae884ce3cebfd7c7c5983e2.

Small question though. What is your reason the favor one over another?

As I understand, Ruby symbols bear the same semantic with Elixir atoms. They both indicates something known to your system. You can argue that your system should know about the `:id` parameter since it comes from Rails router path. So in this case, using symbol is ok.

I wonder about the parameters which users supply:
```
# Should we use string
params['some_wild_parameter']

# Or use symbol here?
params['some_wild_parameter']
```

app/controllers/keywords_controller.rb
Goose97: Fair point 👍 . Fixed in aa34d811ed3ea6cac0f89a19e38db59b5beaea45

andyduong1920 (approved):

app/controllers/keywords_controller.rb
andyduong1920: @Goose97 First let talk about the Rails Parameter; it will convert the Hash params to the symbol key Hash, so either it's the `param from the Rails route` or from the `user supply - through a form` - Rails parameter will convert that, and in the controller, we always can use the `symbol` way -> `params[:some_wild_parameter]`

Back to the reason why we prefer the `symbol` over `string`, first it's just shorter than the quote and makes the code look nice ;) also in Ruby by default - when it see the `string` it will init the new object in the memory and delete it later, while with `symbol` it will just create only 1 single object and keep it there.

config/locales/en.yml
```diff
@@ -8,3 +8,15 @@ en:
       processing: "Processing"
       succeeded: "Succeeded"
       failed: "Failed"
+    shared:
```
olivierobert: Since it is shared, it could moved at the root level:

```
en:
  keywords:
    ...
  shared:
    ...
```

config/locales/en.yml
```diff
@@ -8,3 +8,15 @@ en:
...
+      link:
+        back_to_home: "Back to home page"
+    details:
```
olivierobert: It is best not to map translation keys specifically to pages (as these can change) but by domain. So the following attributes could still fit under `keywords`.

olivierobert (approved):
Looks good :-) Just a few minor improvements ideas.

config/locales/en.yml
Goose97: Fixed in b58d270738575daaea4ef2a0682c10ebc33b0f31

config/locales/en.yml
Goose97: Fixed in 9a6a72b1d07fc729f526a3e0f050536f7879ee2c

PR #40, opened 2023-08-23
[API] As a User, I can get the details of my keywords

### LINKED ISSUE #23: [API] As a User, I can get the details of my keywords
## Why

Provide an API for third-parties to fetch details about a given keyword

## Acceptance Criteria

- The API accepts GET method with the path as: `keywords/:keyword_id`
-  The API MUST return all informations listed in #21
- Do NOT implement authentication. This will be added later once we set up the authentication pipeline

Goose97:
Closes [[API] As a User, I can get the details of my keywords](https://github.com/Goose97/google-scraper-ruby/issues/23)

## What happened 👀

Implement API to get keyword details

## Insight 📝

- API path: `GET keywords/:keyword_id`
- Return all fields listed in #21
- Return 404 if keyword doesn't exist

## Proof Of Work 📹

N/A

### DIFF
--- a/app/serializers/v1/keyword_serializer.rb
+++ b/app/serializers/v1/keyword_serializer.rb
@@ -1,12 +1,10 @@
 # frozen_string_literal: true
 
 module V1
-  class KeywordSerializer < ApplicationSerializer
-    attributes :content, :status, :created_at, :updated_at
+  class KeywordSerializer < V1::KeywordSimpleSerializer
+    attributes :result_page_html, :links_count
 
-    attributes :result_page_html, :links_count, if: proc { |_, params| params[:action] == :show }
-
-    attributes :search_entries_info, if: proc { |_, params| params[:action] == :show } do |_, params|
+    attributes :search_entries_info do |_, params|
       {
         top_ads_count: params[:search_entries_query].top_ads_count,
         total_ads_count: params[:search_entries_query].total_ads_count,
--- a/app/serializers/v1/keyword_serializer.rb
+++ b/app/serializers/v1/keyword_serializer.rb
@@ -3,5 +3,17 @@
 module V1
   class KeywordSerializer < ApplicationSerializer
     attributes :content, :status, :created_at, :updated_at
+
+    attributes :result_page_html, :links_count, if: proc { |_, params| params[:action] == :show }
+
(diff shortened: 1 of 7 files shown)


app/serializers/v1/keyword_serializer.rb
```diff
@@ -3,5 +3,17 @@
 module V1
   class KeywordSerializer < ApplicationSerializer
     attributes :content, :status, :created_at, :updated_at
+
+    attributes :result_page_html, :links_count, if: proc { |_, params| params[:action] == :show }
+
+    attributes :search_entries_info, if: proc { |_, params| params[:action] == :show } do |_, params|
```
olivierobert: Since serializers can be used in other endpoints (via associations), it is best to split the definition of serializer to avoid complexity and unexpected responses:
- 1 serializer for collections
- A serializer for single records 

You can check the following API endpoint implementation in https://github.com/nimblehq/the1-api/blob/development/engines/the1_reward/app/serializers/the1_reward/v2/reward_serializer.rb:
image

Goose97: Fixed in c89ee2f160d3f8f4824711a6bea1b6972417dcfb

app/serializers/v1/keyword_serializer.rb
```diff
@@ -1,7 +1,17 @@
 # frozen_string_literal: true
 
 module V1
-  class KeywordSerializer < ApplicationSerializer
-    attributes :content, :status, :created_at, :updated_at
+  class KeywordSerializer < V1::KeywordSimpleSerializer
+    attributes :result_page_html, :links_count
+
+    attributes :search_entries_info do |_, params|
```
olivierobert: How about placing all these attributes at the root level like `links_count`?

```ruby
attributes :top_ads_count do |_, params|
  params[:search_entries_query].top_ads_count
end

#...
```

It is a bit repetitive but it is a one time cost and it simplifies the response structure 💡

Goose97: Fixed in 94a8558a10fb388a0aede7696441c70800da97ea

andyduong1920 (approved):

olivierobert (approved):

PR #42, opened 2023-08-25
[Backend] As a User, I can search across all uploaded keywords and results

### LINKED ISSUE #25: [Backend] As a User, I can search across all uploaded keywords and results
Backend for #26 

## Why

Provide a tool for users to gain a broader view across all keywords. This will aid users in extracting insights, trends from their keyword data

## Acceptance Criteria

- Support these queries
  - Exact match: i.e. how many times the `apple.com` URL appears? 
  - Partial match: i.e. how many URLs contain the word `ruby`?
  - Pattern match: provide a Regex-subset syntax to perform complex queries
- A query should return how many URLs satisfy the predicate and also what are their corresponding keywords

## Notes

Technically, a three types of queries could be done with the pattern match. However, we reserve pattern matches for only complex queries to not confuse basic users.

Goose97:
Closes [[Backend] As a User, I can search across all uploaded keywords and results](https://github.com/Goose97/google-scraper-ruby/issues/25)

## What happened 👀

Implement backend for scrape result search feature

## Insight 📝

- [x] Implement ScrapeResultSearchQuery. This class is responsible for querying the database and filter URLs against users pattern
  - [x] Perform arguments validation
  - [x] Support different query type
    - [x] exact
    - [x] partial
    - [x] pattern
  - [x] Perform URLs matching in database layer

## Proof Of Work 📹

All tests passed

image


app/services/scrape_result_search_service.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class ScrapeResultSearchService
```
olivierobert: That should not be a service but instead a query object.

Service objects should be reserved for all complex operations except read if the data comes from directly a database.

Goose97: 👍 . Fixed in 0c51983d0128e1c889b63a9fcbe35790f2554c77

olivierobert:
@Goose97 The PR description still has the content placeholder

image

app/queries/scrape_result_search_query.rb
```diff
@@ -0,0 +1,42 @@
...
+        { keyword_id: keyword_id, urls: matched_urls } unless matched_urls.empty?
+      end
+  end
+
+  private
+
+  attr_reader :query_type, :pattern
+
+  def all_search_entries
+    @all_search_entries ||= KeywordSearchEntry.all
```
olivierobert: Loading all data in memory is resource-intensive so it would be best to execute the SQL query to perform matches instead (so not querying all data and filtering after). It would require the use some advanced ActiveRecord methods to eager load associations (search entries) and dive into maybe raw SQL if needed.

spec/queries/scrape_result_search_query_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        Fabricate(:keyword_search_entry, urls: [url_b], keyword_id: keyword.id)
+        Fabricate(:keyword_search_entry, urls: ['https://ruby-lang.org'], keyword_id: keyword.id)
+
+        result = described_class.new(query_type: :pattern, pattern: 'go{2}\w+\.com$').call
+
+        expect(result).to(contain_exactly({ keyword_id: keyword.id, urls: [url_a, url_b] }))
+      end
+    end
+
+    context 'given EMPTY pattern' do
```
olivierobert: Shouldn't these two last tests be in a separate context group, as these are called at the initialize method and complimented with another test?

```ruby
describe '#new' do
  context 'given valid params' do
     it 'does NOT raise errors' do
       #...
     end 
   end 

   context 'given invalid params' do
     context 'given EMPTY pattern' do
       #...
     end 

     context 'given INVALID query_type' do
       #...
     end
   end 
end
```

This is more about having tests self-documenting the implementation and making it easier to understand the implementation

Goose97: Agree. These test cases should belong to a different context. Fixed in f815bca0cc57fccf90353111442612ba1ab7bdc4

app/models/scrape_result_search_params.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class ScrapeResultSearchParams
```
andyduong1920: instead of putting this in the `app/models` we can put it into the `params` folder, like `app/params/..` what do you @Goose97

app/queries/scrape_result_search_query.rb
Goose97: Fixed in 6190314773629b2f8597ae14ce68f8af371ae3db.

I'm not sure whether we should worry about performance for this query, so I went with the most simple one.

But the SQL queries are not very complex, so I guess it's better to go this way.

app/models/scrape_result_search_params.rb
Goose97: Agree. Fixed in 063c7b651f0d62d074ba56a7287bb5e20886a1ff

olivierobert (approved):

andyduong1920 (approved):

PR #43, opened 2023-08-29
[UI] As a User, I can search across all uploaded keywords and results

### LINKED ISSUE #26: [UI] As a User, I can search across all uploaded keywords and results
Frontend for #25 

## Why

Provide a tool for users to gain a broader view across all keywords. This will aid users in extracting insights, trends from their keyword data

## Acceptance Criteria

- Implement a search box to type query input
- Implement a dropdown to select query type
- If the query succeeds, show how many URLs match the given query and their corresponding keywords
- Notify users in case the query encounters any error or timeout

Goose97:
Closes [[UI] As a User, I can search across all uploaded keywords and results](https://github.com/Goose97/google-scraper-ruby/issues/26)

## What happened 👀

Implement UI for scrape result search feature

## Insight 📝

- [x] Implement a search input in keyword index view
- [x] Implement a view to display search result. Path: `keywords/search`
- [x] Redirect to keyword index view if search parameters are invalid

### Users search flow

https://github.com/Goose97/google-scraper-ruby/assets/35915460/81053708-966c-4063-a023-237fbfba0f5f

###  Search result view
image


app/queries/scrape_result_search_query.rb
```diff
@@ -19,16 +19,27 @@ def initialize(pattern:, query_type:)
   end
 
   def call
+    search_entries_filtered_by_url
+      .group(:keyword_id)
+      .includes(:keyword)
```
Goose97, line 24: If I don't want to preload/include all fields from keyword, how can I achieve that 🤔 ?

For example, in this case, I only need the `content` field from the keyword table.

app/assets/stylesheets/screens/_keywords.scss
```diff
@@ -15,6 +15,16 @@ body.keywords {
...
       }
     }
+
+    .scrape-result-search-form {
+      display: flex;
+      justify-content: end;
+      margin-bottom: 1rem;
+
+      &__input {
+        width: 300px;
```
olivierobert: Prefer sizing in `rem`

app/controllers/keywords_controller.rb
```diff
@@ -33,9 +33,33 @@ def create
...
+
+    render(locals: {
+             search_result_presenter: SearchResultPresenter.new(search_result: result),
+             search_params: search_params
+           })
+  rescue ActiveModel::ValidationError => error
+    flash[:alert] = error.model.errors.full_messages
+    redirect_to(keywords_path)
+  end
+  # rubocop:enable Metrics/MethodLength
```
olivierobert, line 51: Disabling Rubcop each time is not great 😅 It might be better to revisit the settings (including for rspect) a bit as this method is not long

app/queries/scrape_result_search_query.rb
olivierobert: You would need to use `.select`:

```suggestion
      .includes(:keyword)
      .select('keyword_search_entry.keyword_id, keyword_search_entry.urls, keyword.content')
```

However, unsure if it is worth it in this case to optimize the queries.

app/controllers/keywords_controller.rb
Goose97: well, the reason I disable Rubocop is this method is actually not long. It exceeds the threshold due to the way I format it. 

IMO, this is a valid case to disable the flag. The code path is linear (no branching) and there're no parts worth extracting out.

app/queries/scrape_result_search_query.rb
Goose97: Thanks. I'm just curious how this be done

app/views/keywords/search.html.erb
```diff
@@ -0,0 +1,21 @@
+<div>
+  <div class="mb-2">
+    <p data-testid="search_result_search_term">Search term: <%= search_params[:search] %></p>
+    <p data-testid="search_result_query_type">Query type: <%= search_params[:query_type] %></p>
+  </div>
+
+  <p data-testid="search_result_total_matched_urls">Total matched URLs: <%= search_result_presenter.total_urls %></p>
```
andyduong1920: Let's use `I18n.t` for `Total matched URLs` and other place @Goose97

app/views/keywords/search.html.erb
```diff
@@ -0,0 +1,21 @@
...
+
+  <% search_result_presenter.search_result.each do |item| %>
+    <div data-keyword-id="<%= item[:keyword_id] %>">
+      <span data-testid="search_result_keyword_content">
+        Keyword content: <%= item[:keyword_content] %>
```
andyduong1920: just a suggestion, should we add the link to the `keywords/:id` to let the user to view that keyword details

spec/systems/keywords/search_scrape_result_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+        select(I18n.t('keywords.query_type.partial'), from: 'query_type')
+
+        click_button I18n.t('shared.button.search')
+      end
+
+      page.assert_selector("[data-testid='search_result_search_term']", text: 'ruby', count: 1)
+      page.assert_selector("[data-testid='search_result_query_type']", text: 'partial', count: 1)
+      page.assert_selector("[data-testid='search_result_total_matched_urls']", text: '2', count: 1)
+
+      within("div[data-keyword-id='#{keyword.id}']") do |item|
```
andyduong1920: let's improve the test data by creating 2 keywords that had the ruby term there @Goose97 so in this test, we can see 2 `within("div[data-keyword-id='#{keyword.id}']") do |item|` blocks mean there is 2 keywords will be displaying, what do you think?

Goose97: Yep, that's better 👍 . Fixed in 15223d84800492f1918f059b6096daf122eaebc9

app/views/keywords/search.html.erb
Goose97: Fixed in 4bdc770c852a0e732c6ea2f7ac897c4a36818a41

app/views/keywords/_scrape_result_search_form.erb
```diff
@@ -0,0 +1,24 @@
...
+      [I18n.t('keywords.query_types.partial'), :partial],
+      [I18n.t('keywords.query_types.pattern'), :pattern]
+    ]),
+    class: 'p-2'
+  ) %>
+  <%= text_field_tag(
+    :search,
+    '',
+    class: 'scrape-result-search-form__input',
+    placeholder: I18n.t('keywords.search_input_placeholder')
```
andyduong1920, line 23: We can add the `required: true` tothis `text_field` so the form won't be submit if the field is empty @Goose97

app/views/keywords/search.html.erb
```diff
@@ -0,0 +1,23 @@
+<div>
+  <div class="mb-2">
+    <p data-testid="search_result_search_term"><%= I18n.t('keywords.search_term') %>: <%= search_params[:search] %></p>
+    <p data-testid="search_result_query_type"><%= I18n.t('keywords.query_type') %>: <%= search_params[:query_type] %></p>
+  </div>
+
+  <p data-testid="search_result_total_matched_urls">
+    <%= I18n.t('keywords.total_matched_urls') %>: <%= search_result_presenter.total_urls %>
```
andyduong1920: @Goose97 just FYI, we can avoid the `:` by using the I18n feature like

```
I18n.t('keywords.total_matched_urls', total_urls: search_result_presenter.total_urls)
```

In the YML file, we will define it

```
total_matched_urls: 'Total matched URLs %{total_urls}'
```

app/assets/stylesheets/screens/_keywords.scss
andyduong1920: @Goose97 if you not sure about this comment, you just need to change this to `rem(300px)`, the magic will happen in the https://github.com/Goose97/google-scraper-ruby/blob/4bdc770c852a0e732c6ea2f7ac897c4a36818a41/app/assets/stylesheets/functions/_sizing.scss#L17 file :D

Goose97: Thanks for the pointer Andy. Fixed in fcf27261f43297c2f70df35a02d3223d925eaf3a

app/views/keywords/search.html.erb
Goose97: TIL 🙇 . Fixed in 91a4eef8138475aa7dfa586f30fbd1f45b1dbc04

app/views/keywords/_scrape_result_search_form.erb
Goose97: Good catch. Fixed in 1cf4e0b6a57c486049f3ec6baae7f1fbb404cb7b

app/controllers/keywords_controller.rb
olivierobert: The issue is that there are numerous such disabling in the codebase 😅 I agree that the code is not too long or complex so it would be best to check if the current limit makes sense for this project.

olivierobert (approved):

app/controllers/keywords_controller.rb
Goose97: Totally agree on that. Let me revise the rubocop configuration

app/views/keywords/search.html.erb
Goose97: Fixed in 67326c06c9e333e39f71bc799260469d1f75f8dc

PR #49, opened 2023-09-06
[API] As a User, I can upload CSV keyword files

### LINKED ISSUE #24: [API] As a User, I can upload CSV keyword files
## Why

Third-party applications needs an interface to upload keywords to our system

## Acceptance Criteria

- Provide an API to upload keyword files
- Keyword files should be validated. Rules are specified in #9 and return 4xx error if the validation fails
- When succeeds, the API should response with a hyperlink. Such hyperlink could be polled continuously to keep track of the progress

Goose97:
Closes [[API] As a User, I can upload CSV keyword files](https://github.com/Goose97/google-scraper-ruby/issues/24)

## What happened 👀

Implement CSV files upload API

## Insight 📝

- [x] IMPORTANT: CsvUploadForm will now include the newly created keywords in the form object after successfully processed
- [x] Implement CSV files upload at `POST api/v1/keywords`

## Proof Of Work 📹

image

### DIFF
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -32,11 +32,30 @@ module Api
         render(json: ::V1::SearchResultSimpleSerializer.new(result, is_collection: true))
       end
 
+      def create
+        form = CsvUploadForm.new
+
+        if form.save(create_params[:file])
+          render(
+            json: ::V1::KeywordSimpleSerializer.new(form.keywords, is_collection: true),
+            status: :created
+          )
+        else
+          render_error(status: :unprocessable_entity, code: :invalid_csv_file)
+        end
+      end
+
+      private
+
       def search_params
         {
           search: params.require(:search),
           query_type: params.require(:query_type).to_sym
         }
+      end
+
+      def create_params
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/application_controller.rb
```diff
@@ -23,8 +23,6 @@ class ApplicationController < ActionController::API
         )
       end
 
-      private
```
Goose97: This is now a public API. It will be use directly in controller. For example, [this](https://github.com/Goose97/google-scraper-ruby/pull/49/files#diff-7d00cc43aaf2e771dd8efea4a52134720dfdca8b73776eca01e7c55fd6408457R30)

olivierobert: There seems to be a misunderstanding as having this method set as private does not prevent calling it from a subclass. See https://stackoverflow.com/questions/1565269/are-there-good-reasons-for-private-to-work-the-way-it-does-in-ruby/1565640#1565640

Goose97: Oh, I see. Thanks for pointing it out. Rebased and fixed

olivierobert (approved):

PR #50, opened 2023-09-08
[Backend][UI] As a User, I can sign up with username and password

### LINKED ISSUE #4: [Backend] As a User, I can sign up with username and password
Backend for #5 

## Why
The application requires user authentication to correctly separate data between different users. To achieve that, users must have a way to register themselves to the application.

## Acceptance Criteria
- Users must be able to register an account
- Upon successful registration, users should be directed to the dashboard page
- Upon unsuccessful registration, users should be notify with the failure reason

### LINKED ISSUE #5: [UI] As a User, I can sign up with username and password
Frontend for #4 

## Why
The application requires user authentication to correctly separate data between different users. To achieve that, users must have a way to register themselves to the application.

## Acceptance Criteria

- Implement a form with two attributes: `email` and `password`
- Enable the submit button only if both inputs are filled (optional)
- Show validation error in case of authentication error.

Goose97:
Closes [[Backend] As a User, I can sign up with username and password](https://github.com/Goose97/google-scraper-ruby/issues/4)
Closes [[UI] As a User, I can sign up with username and password](https://github.com/Goose97/google-scraper-ruby/issues/5)

## What happened 👀

Implement sign up feature using Devise gem

## Insight 📝

IMPORTANT: this PR does not integrate authentication into existing features (view, upload, ... keywords). It will be address by another story

- [x] Add sign up page at path `users/sign_up`
- [x] Redirect users to root path if sign-up is successful
- [x] Notify users with error if sign-up is failed

## Proof Of Work 📹

https://github.com/Goose97/google-scraper-ruby/assets/35915460/57f08e91-51cd-403f-8286-b7758a840f91


spec/systems/users/view_sign_up_page_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'View sign up page', type: :system do
```
Goose97, line 5: While it seems like we're testing Devise implementation, these tests still offer some values. They establish our requirements, safe-guard future developments, i.e. customize sign up view

app/helpers/alert_helper.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module AlertHelper
```
olivierobert: Since it is a shared concern, it could be in that application helper. There is not much else we would add to that module moving forward 😅

app/views/devise/registrations/new.html.erb
```diff
@@ -0,0 +1,29 @@
+<h2>Sign up</h2>
+
+<%= form_for(resource, as: resource_name, url: registration_path(resource_name)) do |f| %>
```
olivierobert: The "newer" helper is [form_with](https://guides.rubyonrails.org/form_helpers.html). `form_for` is the older, still correct though, helper.

config/initializers/devise.rb
```diff
@@ -0,0 +1,313 @@
...
+
+  # ==> Controller configuration
+  # Configure the parent class to the devise controllers.
+  # config.parent_controller = 'DeviseController'
+
+  # ==> Mailer Configuration
+  # Configure the e-mail address which will be shown in Devise::Mailer,
+  # note that it will be overwritten if you use your own mailer class
+  # with default "from" parameter.
+  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
```
olivierobert: This requires to be edited

config/initializers/devise.rb
```diff
@@ -0,0 +1,313 @@
...
+#
+# Use this hook to configure devise mailer, warden hooks and so forth.
+# Many of these configuration options can be set straight in your model.
+Devise.setup do |config|
+  # The secret key used by Devise. Devise uses this key to generate
+  # random tokens. Changing this key will render invalid all existing
+  # confirmation, reset password and unlock tokens in the database.
+  # Devise will use the `secret_key_base` as its `secret_key`
+  # by default. You can change it below and use your own secret key.
+  # config.secret_key = '3b7592ae59968d71cc7eb4a4469c8529dbd072ab085bf3ec55f754924fd922ac5e55b9790e28769ac818cbcb6fab6535e5d10779889a79d4c966cd1c92b4914b'
```
olivierobert: This requires to be defined and an env variable must be defined for it

db/migrate/20230908042751_devise_create_users.rb
```diff
@@ -0,0 +1,44 @@
...
+      # t.datetime :confirmed_at
+      # t.datetime :confirmation_sent_at
+      # t.string   :unconfirmed_email # Only if using reconfirmable
+
+      ## Lockable
+      # t.integer  :failed_attempts, default: 0, null: false # Only if lock strategy is :failed_attempts
+      # t.string   :unlock_token # Only if unlock strategy is :email or :both
+      # t.datetime :locked_at
+
+
```
olivierobert: All of the above + the extra line break can be removed as unused.

spec/systems/users/view_sign_up_page_spec.rb
olivierobert: Thank for teh clarification

app/views/devise/registrations/new.html.erb
Goose97: Thanks for the tip. Fixed in 0f4884868c052dc7399bd767129fc98cdf46200e

config/initializers/devise.rb
Goose97: Where is the preferred ways to set environment variables?

1. In the `.env` file
2. In the `config/application.yml` file

app/helpers/alert_helper.rb
Goose97: 👍 . Fixed in 333c9310fb5607fd102affc3a065b388338c98d9

config/initializers/devise.rb
olivierobert: `application.yml` comes with the dependency [figjam](https://index.ruby-china.com/gems/figjam) so that should be used. There is ongoing effort to move to `.env` file in the Rails template but that is not done yet.

Goose97: Fixed in 1ef1d9f7f9618860ae2b904c4aba9e23e6eb0c72

db/migrate/20230908042751_devise_create_users.rb
Goose97: Fixed in f12e435cf474529f25e809f5e8d0f33123dd5c70

config/initializers/devise.rb
Goose97: Fixed in 1ef1d9f7f9618860ae2b904c4aba9e23e6eb0c72

config/initializers/devise.rb
```diff
@@ -0,0 +1,313 @@
...
+#
+# Use this hook to configure devise mailer, warden hooks and so forth.
+# Many of these configuration options can be set straight in your model.
+Devise.setup do |config|
+  # The secret key used by Devise. Devise uses this key to generate
+  # random tokens. Changing this key will render invalid all existing
+  # confirmation, reset password and unlock tokens in the database.
+  # Devise will use the `secret_key_base` as its `secret_key`
+  # by default. You can change it below and use your own secret key.
+  config.secret_key = ENV['DEVISE_SECRET_KEY']
```
olivierobert: For a production-ready app, it is best to fail the boot-up of the app if such a critical env is missing. So prefer:

```suggestion
  config.secret_key = ENV.fetch'DEVISE_SECRET_KEY']
```

See https://stackoverflow.com/a/47985757

olivierobert (approved):
One minor improvement so pre-approving it

app/views/shared/_alert.erb
```diff
@@ -2,7 +2,7 @@
   <div>
     <% if flash[:alert] %>
       <div class="alert alert-danger" data-testid="alert-error">
-        <% flash[:alert].each do |error| %>
+        <% flash_alert.each do |error| %>
```
andyduong1920: How about 
```suggestion
        <% flash_alerts.each do |error| %>
```

config/application.yml
```diff
@@ -13,6 +13,7 @@ default: &default
 
 development:
   <<: *default
+  DEVISE_SECRET_KEY: "6a8b67c33f0cd9be1a41ab9989bcceeb1cc46916dc494b105092513efb8c7d2e3c5fc04bd88b6fc0163c0dcbc141a5e65effbfe78fc65090a5cf4fcaf06179b9"
```
andyduong1920: it's ok to have the same key for both development and test @Goose97 so I prefer to move this into the `default` scope above :)

spec/systems/users/sign_up_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+describe 'Sign up', type: :system do
+  context 'given VALID parameters' do
+    it 'logs user in and redirects to the root path' do
+      visit(new_user_registration_path)
+      fill_in('user[email]', with: 'abc@gmail.com')
+      fill_in('user[password]', with: '123456')
+      fill_in('user[password_confirmation]', with: '123456')
+      click_button(I18n.t('auth.sign_up'))
+
+      expect(page).to(have_current_path(root_path))
```
andyduong1920, line 14: Do we have any notice on this screen like `Sign up successfully` message, if yes, we can expect it is displaying here, what do you think @Goose97

app/views/shared/_alert.erb
Goose97: Agree. Fixed in dec1bad0279edfaebcefdb49cfa554f85aa47e82

config/application.yml
Goose97: Fixed in ad0b819b7bf8209387ac7547aded354e6c37f039

spec/systems/users/sign_up_spec.rb
Goose97: Fixed in e0412cf20118877470bb72ca1c246319d835d62f

andyduong1920 (approved):

Issue #56, opened 2023-09-20
[Backend] As a User, I can only see my keywords


Goose97:
## Why

Users must be authenticated before using the app, so that we can segregate users data. An user can only see and interact their uploaded keywords. 

## Acceptance Criteria

- [x] For an unauthenticated user, redirect them to the sign in page
- [x] Add user_id to the keywords table

Affected features:
  - [x] List keywords
  - [x] See keyword details
  - [x] Search keywords result

PR #57, opened 2023-09-20
[Backend] As a User, I can only see my keywords

Goose97:
[[Backend] As a User, I can only see my keywords](https://github.com/Goose97/google-scraper-ruby/issues/56)

## What happened 👀

Integrate authentication into upload keywords and view keywords features

## Insight 📝

- [x] Add user_id column to `keywords` table and assocications
- [x] Use user_id when create new keywords
- [x] Only returns user keywords when visit `keywords#index`

Some API features is broken at this point, that's why I [skipped some tests](https://github.com/Goose97/google-scraper-ruby/pull/57/files#diff-0310621c19316fa223a044dfa0ccc30f303558664a4933677118123aeba687ceR112). It will be handled in https://github.com/Goose97/google-scraper-ruby/issues/59

## Proof Of Work 📹

https://github.com/Goose97/google-scraper-ruby/assets/35915460/8f2481ac-3511-4cf2-9a60-b49c62ac8b5f


Goose97:
I wonder where to put the test case:

```
context 'given a authenticated user' do
  it "returns only user's keywords" do
  end
end
```

Should it be in the controller test or the query object test?

app/models/keyword_search_entry.rb
```diff
@@ -7,4 +7,6 @@ class KeywordSearchEntry < ApplicationRecord
   belongs_to :keyword
 
   validates :kind, :position, :urls, presence: true
+
+  scope :belongs_to_user, ->(user_id) { joins(:keyword).where(keywords: { user_id: user_id }) }
```
olivierobert, line 11: It see best to use the gem pundit for that purpose as thiw will get repetitive. See the following:
- Pundit Policy: https://github.com/nimblehq/mia-web/blob/development/app/policies/campaign_policy.rb (especially the scope part which handles the query)
- Controller usage: https://github.com/nimblehq/mia-web/blob/development/engines/mia_campaign/app/controllers/mia_campaign/campaigns_controller.rb#L9

app/controllers/keywords_controller.rb
```diff
@@ -2,7 +2,8 @@
 
 class KeywordsController < ApplicationController
   def index
-    pagy, keywords = pagy(KeywordsQuery.new.call)
+    query = KeywordsQuery.new(scope: current_user.keywords)
```
olivierobert, line 5: See below the comment about using Pundit

app/forms/csv_upload_form.rb
```diff
@@ -18,7 +22,7 @@ def save(file)
 
     @keywords = ActiveRecord::Base.transaction do
       keywords.map do |keyword|
-        Keyword.create(content: keyword)
+        user.keywords.create(content: keyword)
```
olivierobert, line 25: While that is possible, the convention is to use the model directly:

```suggestion
        Keyword.create(content: keyword, user: user) 
```

The main reason is that it is will be more consistent in the whole codebase. Usually in the controller, the following code is used:

```
def some_action
   SomeModel.create(some_params)
end

private

def some_params
   params.permit(:param_1, :param_2).merge(user_id: current_user)
end
```

See https://github.com/nimblehq/mia-web/blob/752f45a860b25bdc456c1bcc82b6b5a6ca18e334/engines/mia_payment/app/controllers/mia_payment/bank_accounts_controller.rb

spec/fabricators/keyword_fabricator.rb
```diff
@@ -3,6 +3,7 @@
 Fabricator(:keyword) do
   content { FFaker::Lorem.characters(16) }
   status :processing
+  user
```
olivierobert, line 6: Would this do the same as the following:

```suggestion
  user { Fabricate(:user) }
```

app/queries/scrape_result_search_query.rb
```diff
@@ -10,10 +10,11 @@ class ScrapeResultSearchQuery
     pattern: 'url ~ ?'
   }.freeze
 
-  def initialize(pattern:, query_type:)
+  def initialize(pattern:, query_type:, scope: KeywordSearchEntry)
```
olivierobert, line 13: I am unsure why this change and the one for `KeywordsQuery` are needed. Please explain.

andyduong1920:
@Goose97 Rebase required 🙏

andyduong1920:
Rebase required @Goose97 👍

